PR #38, opened 2021-01-27
fix(modelgen-android): add additional param to query fields

### LINKED ISSUE #21: Android - Add additional parameter to generated model's query fields
Relevant issue: https://github.com/aws-amplify/amplify-android/issues/908

There is currently an issue with the fact that model query fields only store the field name, which makes it ambiguous when there are multiple models containing same field names while querying from SQLite database (e.g. "get Post where name is _x_" can be ambiguous whether you are filtering by Post's name or Blog's name, if Posts have "belongs to" relationship with Blogs).

This issue can be addressed by storing the model name in query field instance too (https://github.com/aws-amplify/amplify-android/pull/1133) to prevent ambiguity when referring to field names.

Java models for Android are currently generated with static constants using old `QueryField` instance with just the field name. To allow customers to reliably use these constants, the models should be generated with query fields that specify both the model's name and field name.

**Current codegen behavior:**
```
public final class Blog implements Model {
  public static final QueryField ID = field("id");
  public static final QueryField NAME = field("name");
  public static final QueryField OWNER = field("blogOwnerId");
  //...
```

**Proposed codegen behavior:**
```
public final class Blog implements Model {
  public static final QueryField ID = field("Blog", "id");
  public static final QueryField NAME = field("Blog", "name");
  public static final QueryField OWNER = field("Blog", "blogOwnerId");
  //...
```

Note: pending release of Android Amplify v1.6.11

AaronZyLee:
_Issue #, if available:_
fix #21 
_Description of changes:_
Add the model name to query field to prevent ambiguity when referring to field names from multiple models
```Java
public final class Blog implements Model {
  public static final QueryField ID = field("Blog", "id");  //used to be 'field("id")'
  //...
}
```
`jest java-visitor`

### DIFF
--- a/packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-java-visitor.test.ts.snap
+++ b/packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-java-visitor.test.ts.snap
@@ -27,9 +27,9 @@ import static com.amplifyframework.core.model.query.predicate.QueryField.field;
   @AuthRule(allow = AuthStrategy.OWNER, ownerField = \\"owner\\", identityClaim = \\"user_id\\", operations = { ModelOperation.CREATE, ModelOperation.UPDATE, ModelOperation.DELETE, ModelOperation.READ })
 })
 public final class customClaim implements Model {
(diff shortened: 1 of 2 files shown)


raphkim (approved):
🚢

PR #74, opened 2021-02-17
TypeGen(swift): Generate S3ObjectInput related extensions only if all required fields are present

phani-srikar:
_Issue #, if available:_
#52 

_Description of changes:_
These changes fix the issue with swift types generation where conformance to `AWSS3InputObjectProtocol` extension was generated even when the `localUri` and `mimeType` fields are either optional or not present in the schema S3Object. This caused build errors for iOS projects. 

unit test included for changes, run using `jest`


AaronZyLee (approved):
Just want to confirm that both `localUri` and `mimeType` should be non-nullable for using AWSAppsycn iOS SDK. Otherwise LGTM 🎉

PR #77, opened 2021-02-18
fix(amplify-codegen-appsync-model-plugin): include isNullable & isList in version hash

amhinson:
*Issue #, if available:*
Given a basic schema in a JavaScript project:
```graphql
type Post @model {
  id: ID!
  title: String
}
```
after running `amplify models codegen` for the first time, the `schema.js` file is populated as expected with a unique `version` hash based on the type and the fields.

However, after changing `title` to be required with `String!`, and then running `amplify models codegen` again, the `version` doesn't change.

The same behavior also applies when changing a field from `String` to `[String]`.

My expectation is that a field's `isNullable` & `isList` attribute should be considered in this hash.

*Description of changes:*
- include `isNullable` & `isList` when calculating the `version` hash.

- tested locally with a sample app as described above.

---

### DIFF
--- a/packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
+++ b/packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
@@ -392,6 +392,7 @@ export class AppSyncModelVisitor<
             directives: fieldDirectives,
             type: field.type,
             isNullable: field.isNullable,
+            isList: field.isList,
           };
         })
         .sort((a, b) => sortFields(a, b));
--- a/packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
+++ b/packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
@@ -391,6 +391,7 @@ export class AppSyncModelVisitor<
             name: field.name,
             directives: fieldDirectives,
             type: field.type,
+            isNullable: field.isNullable,
           };
         })
         .sort((a, b) => sortFields(a, b));
--- a/packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
+++ b/packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
@@ -393,6 +393,7 @@ export class AppSyncModelVisitor<
             type: field.type,
             isNullable: field.isNullable,
             isList: field.isList,
+            isListNullable: field.isListNullable,
           };
         })
         .sort((a, b) => sortFields(a, b));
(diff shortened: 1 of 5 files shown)


AaronZyLee (commented):
Is the version change applied to only JS or for all platforms? From the test result it also impacts dart and swift code. If this is the case, the PR should need additional review from iOS and flutter team.

amhinson:
It appears to impact all platforms, since they all utilize the `version`. I imagine that they all would need to know if this sort of schema change happened.

Amplifiyer:
The change looks good! Maybe we should add another unit test that changes the model schema and verify that the version has changed?

Also will it take care of changing between list of nullable items vs required items? `intList: [Int!]` vs `intList: [Int]`

amhinson:
> The change looks good! Maybe we should add another unit test that changes the model schema and verify that the version has changed?
> 
> Also will it take care of changing between list of nullable items vs required items? `intList: [Int!]` vs `intList: [Int]`

Good callouts 👍 

I've added some unit tests around this as well as the use-case that you pointed out. That particular case did work as expected, but that also made me realize that we should include `isList` in the hash for cases when a field switches from `String` to `[String]`.

Amplifiyer (approved):

phani-srikar (approved):
LGTM! Thanks for updating the tests.

packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
```diff
@@ -391,6 +391,8 @@ export class AppSyncModelVisitor<
             name: field.name,
             directives: fieldDirectives,
             type: field.type,
+            isNullable: field.isNullable,
+            isList: field.isList,
```
AaronZyLee, line 395: I think there is one scenario missing for nullability for list. Suppose in the two schema the only difference is field `[String]` and `[String]!`. And this will have the same computed version. I suggest to add an additional line as following and related unit tests.
```suggestion
            isList: field.isList,
            isListNullable: field.isListNullable
```

AaronZyLee (changes requested):

packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
AaronZyLee: FYI `isListNullable` is an attribute for field after loading https://github.com/aws-amplify/amplify-codegen/blob/master/packages/appsync-modelgen-plugin/src/utils/get-type-info.ts

amhinson: Nice catch! I just added this with a corresponding test 👍

AaronZyLee (approved):
🚢

PR #76, opened 2021-02-17
Migrate amplify-codegen package

AaronZyLee:
_Issue #, if available:_

_Description of changes:_
Migrate `amplify-codegen` from cli to codegen repo.
Bump the version to the next minor version: **2.22.0**
_How are these changes tested:_

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #78, opened 2021-02-18
Migrate amplify-codegen package (#76)

AaronZyLee:
Release the migrated amplify-codegen


AaronZyLee:
Close it since the `[ci skip]` tags in commit

PR #84, opened 2021-02-20
feat(modelgen): fetch graphql schemas from nested input folder

phani-srikar:
_Issue #, if available:_.  #75 

_Description of changes:_
These changes enable developers to organize their GraphQL schema files in nested directories eliminating the hard constraint to have all schema files at the root of `schema` folder. 

Note:
If the developer however, also has a `schema.graphql` file in the path as specified in prompt `Edit your schema at project/typescriptapp/amplify/backend/api/typescriptapp/schema.graphql or place .graphql files in a directory at project/typescriptapp/amplify/backend/api/typescriptapp/schema`, that takes precedence and the `schema` folder is not searched. 

- using unit tests
- using sample typescript and flutter apps

### DIFF
--- a/packages/amplify-codegen/src/commands/models.js
+++ b/packages/amplify-codegen/src/commands/models.js
@@ -1,6 +1,7 @@
 const path = require('path');
 const fs = require('fs-extra');
 const { parse } = require('graphql');
+const glob  = require('glob-all');
 const { FeatureFlags, pathManager } = require('amplify-cli-core');
 const gqlCodeGen = require('@graphql-codegen/core');
 const { getModelgenPackage } = require('../utils/getModelgenPackage');
@@ -104,12 +105,11 @@ function loadSchema(apiResourcePath) {
     return fs.readFileSync(schemaFilePath, 'utf8');
   }
   if (fs.pathExistsSync(schemaDirectory) && fs.lstatSync(schemaDirectory).isDirectory()) {
-    return fs
-      .readdirSync(schemaDirectory)
-      .map(file => path.join(schemaDirectory, file))
-      .filter(file => file.endsWith('.graphql') && fs.lstatSync(file).isFile())
-      .map(file => fs.readFileSync(file, 'utf8'))
-      .join('\n');
+    // search recursively for graphql schema files inside `schema` directory
+    const schemas = glob.sync([
+      path.join(schemaDirectory, '**/*.graphql')
+    ]);
+    return schemas.map(file => fs.readFileSync(file, 'utf8')).join('\n');
   }
 
   throw new Error('Could not load the schema');
--- a/packages/amplify-codegen/tests/commands/models.test.js
+++ b/packages/amplify-codegen/tests/commands/models.test.js
@@ -109,5 +109,5 @@ describe('command-models-generates models in expected output path', () => {
         });
     };
 
-    afterEach(mock.restore);
+    afterEach(mockFs.restore);
   });
--- a/packages/amplify-codegen/tests/commands/models.test.js
+++ b/packages/amplify-codegen/tests/commands/models.test.js
@@ -1,0 +1,113 @@
+const generateModels = require('../../src/commands/models');
+const mock = require('mock-fs');
+const graphqlCodegen = require('@graphql-codegen/core');
+const fs = require('fs');
+const path = require('path');
+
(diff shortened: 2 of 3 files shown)


packages/amplify-codegen/src/commands/models.js
```diff
@@ -104,12 +105,11 @@ function loadSchema(apiResourcePath) {
...
   if (fs.pathExistsSync(schemaDirectory) && fs.lstatSync(schemaDirectory).isDirectory()) {
-    return fs
-      .readdirSync(schemaDirectory)
-      .map(file => path.join(schemaDirectory, file))
-      .filter(file => file.endsWith('.graphql') && fs.lstatSync(file).isFile())
-      .map(file => fs.readFileSync(file, 'utf8'))
-      .join('\n');
+    // search recursively for graphql schema files inside `schema` directory
+    const schemas = glob.sync([
+      path.join(schemaDirectory, '**/*.graphql')
```
AaronZyLee, line 110: There is a `build/` folder under the `schemaDirectory` for amplify project, which has the built graphql files. Those files are not the input for model-gen and should be excluded in the glob expression. I think an additional line should be applied in glob expression: `!${schemaDirectory/build/**/*.graphql`

AaronZyLee (changes requested):

packages/amplify-codegen/src/commands/models.js
phani-srikar: The `schemaDirectory` here points to `project/typescriptapp/amplify/backend/api/typescriptapp/schema` as specified in the modelgen prompt. `build/` folder is out of scope for this folder and glob expression as it is one level above `schema` folder.

packages/amplify-codegen/tests/commands/models.test.js
```diff
@@ -0,0 +1,113 @@
...
+                    }
+                ]
+            }
+        );
+      MOCK_CONTEXT.amplify.executeProviderUtils.mockReturnValue([]);
+      MOCK_CONTEXT.amplify.pathManager.getBackendDirPath.mockReturnValue(MOCK_BACKEND_DIRECTORY);
+      graphqlCodegen.codegen.mockReturnValue(MOCK_GENERATED_CODE);
+    });
+
+    for (const frontend in OUTPUT_PATHS) {
```
AaronZyLee, line 70: Do we need the for loop for all frontend types? I think the tests can be simplified to one by mocking a frontend and its output dir and check if codegen function gets called.

packages/amplify-codegen/tests/commands/models.test.js
```diff
@@ -0,0 +1,113 @@
+const generateModels = require('../../src/commands/models');
+const mock = require('mock-fs');
```
AaronZyLee, line 2: It is better to change the importing name to `mockFs`(or others) to avoid confusion

AaronZyLee (changes requested):

packages/amplify-codegen/tests/commands/models.test.js
```diff
@@ -0,0 +1,113 @@
...
+    getEnvInfo: jest.fn(),
+    getResourceStatus: jest.fn(),
+    executeProviderUtils: jest.fn(),
+    pathManager: {
+        getBackendDirPath: jest.fn()
+    },
+    getProjectConfig: jest.fn()
+  },
+};
+const OUTPUT_PATHS = {
```
AaronZyLee, line 23: The constants here is not linked to the source code and will break if the default path changes. I suggest to use mock frontend and mock ouput path.

packages/amplify-codegen/tests/commands/models.test.js
phani-srikar: The function `getModelOutputPath` is not exported currently and is a private function, which prevents us from mocking it. The for loop was intended to avoid code duplication and test the model generation (mainly `loadSchema` and `getModelOutputPath`) for all front-ends.

packages/amplify-codegen/tests/commands/models.test.js
phani-srikar: Yes. There are the expected output_paths where models should be generated for each frontend. If a change is made in the source files to change this behavior, it can be caught here in which case we will update the test too with new output path.

AaronZyLee (approved):
🚢

Issue #50, opened 2020-11-24
iOS/codegen: "generated" models folder and definitions go out of sync after updating schema.graphql


diegocstn:
**Describe the bug**
Changing a model's name or removing a model from `schema.graphql` is not reflected in `amplify/generated/models/`.

**Amplify CLI Version**
4.35.0

**To Reproduce**
- Follow iOS [getting started guide ](https://docs.amplify.aws/lib/datastore/getting-started/q/platform/ios) to setup a new iOS project
- run `amplify codegen models`
- run `amplify-app`
- **rename** or **delete** one of the entity in the `schema.graphql` schema
- run `amplify codegen models` and `amplify-app`
- observe that the generated models don't reflect the current schema status, removed/renamed model is still present

**Expected behavior**
The renamed/removed model gets also renamed/removed from both the `amplify/generated/models` folder Xcode project

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - Mac
 - v12.12.0

**Additional context**
N/A

PR #85, opened 2021-02-22
feat(modelgen): support clearing outdated models using a feature flag

phani-srikar:
_Issue #, if available:_. #50 

_Description of changes:_
These changes allow the developers to automatically remove outdated model files upon schema update. 
These will be released with a Feature Flag so that existing customers will not have any breaking changes.

using unit tests and TS, Dart sample apps. 


packages/amplify-codegen/tests/commands/models.test.js
```diff
@@ -71,42 +71,128 @@ describe('command - models', () => {
...
+            const outputDirectory = path.join(MOCK_PROJECT_ROOT, OUTPUT_PATHS[frontend]);
+            const preExistingFileName = 'preexisting.txt';
+            const preExistingFilePath = path.join(outputDirectory, preExistingFileName);
+            const mockedFiles = {};
+            mockedFiles[schemaFilePath] = {
+                'schema.graphql': ' type SimpleModel { id: ID! status: String } '
+            };
+            mockedFiles[outputDirectory] = {};
+            mockedFiles[outputDirectory][preExistingFileName] = 'A pre-existing file to be cleared';
+            mock(mockedFiles);
```
AaronZyLee, line 169: Could you rebase this PR and change the import to `mockFs` as the previous review?

packages/amplify-codegen/tests/commands/models.test.js
```diff
@@ -71,42 +71,128 @@ describe('command - models', () => {
...
+
+            // assert model generation succeeds
+            expect(graphqlCodegen.codegen).toBeCalled();
+
+            // assert that codegen generated in correct place
+            expect(fs.readdirSync(outputDirectory).length).toBeGreaterThan(1);
+
+            // assert that the pre-existing file is retained
+            expect(fs.existsSync(preExistingFilePath)).toBeTruthy;
+        });
```
AaronZyLee, line 195: These two unit tests have some duplicate code at the beginning. You could use a new test suite and `beforeEach` to include the duplicate parts. And also there is no need to loop all frontend types for clearing tests.

AaronZyLee (changes requested):

packages/amplify-codegen/tests/commands/models.test.js
phani-srikar: Moved these to a separate test suite and refactored common mock setup to a method. Retained the tests for all frontends, as they have different output paths.

packages/amplify-codegen/tests/commands/models.test.js
```diff
@@ -71,43 +72,148 @@ describe('command - models', () => {
...
+      expect(fs.readdirSync(outputDirectory).length).toEqual(1);
+      expect(fs.existsSync(preExistingFilePath));
+
+      await generateModels(MOCK_CONTEXT);
+
+      // assert model generation succeeds
+      expect(graphqlCodegen.codegen).toBeCalled();
+
+      // assert that codegen generated in correct place
+      expect(fs.readdirSync(outputDirectory).length).toBeGreaterThan(1);
```
AaronZyLee, line 199: The previous test set the value to 0 but this is 1 instead. Is this intended change?

packages/amplify-codegen/tests/commands/models.test.js
```diff
@@ -71,43 +72,148 @@ describe('command - models', () => {
...
+                    service: 'AppSync', 
+                    providerPlugin: 'awscloudformation',
+                    resourceName: MOCK_PROJECT_NAME
+                  }
+              ]
+          }
+      );
+    MOCK_CONTEXT.amplify.executeProviderUtils.mockReturnValue([]);
+    MOCK_CONTEXT.amplify.pathManager.getBackendDirPath.mockReturnValue(MOCK_BACKEND_DIRECTORY);
+    graphqlCodegen.codegen.mockReturnValue(MOCK_GENERATED_CODE);
```
AaronZyLee, line 141: The beforeEach in two test suites(line52-69 and line125-142) seems to have the same code/mock values. Can you move the duplicate mocking to the beginning?

AaronZyLee (changes requested):

packages/amplify-codegen/tests/commands/models.test.js
phani-srikar: Refactored it to a method. Thanks

packages/amplify-codegen/tests/commands/models.test.js
phani-srikar: This one tests `Should not clear the output directory before generation if cleanGeneratedModelsDirectory is false` and hence the pre-existing file should not be deleted.

AaronZyLee (approved):
LGTM 🐐

PR #98, opened 2021-03-03
feat(documentation): add a guide to migration related Feature Flags

phani-srikar:
_Description of changes:_
Adds a guide to better explain the package migration related Feature Flags.
https://docs.amplify.aws/cli/reference/feature-flags#codegen

### DIFF
--- a/FeatureFlags.md
+++ b/FeatureFlags.md
@@ -16,7 +16,9 @@ In order to shield the existing customers from any complications as a result of 
 The package implementations in this repository will be actively maintained going forward while the implementations in [Amplify CLI repository](https://github.com/aws-amplify/amplify-cli/tree/master/packages) will be deprecated on May 1st, 2021. 
 The package implementations in this repository will support all the APIs that are supported by their counterparts in the CLI repository. 
 
-Hence, we recommend setting the above Feature Flags to `true` in your `cli.json` file to consume the package implementations from this repository and take advtange of bug fixes, enhancements etc being made to these generators.
+These Feature Flags are set to `true` by default for new projects created using Amplify CLI and `false` for existing projects.
+We recommend setting the above Feature Flags to `true` in your `cli.json` file to consume the package implementations from this repository and take advtange of bug fixes, enhancements etc being made to these generators.
+
 Some examples of recent enhacements include, `cleanGeneratedModelsDirectory` and `retainCaseStyle` features that are only supported when the above Feature Flags are set to true. 
 Please refer to the individual package changelogs in this repository for a comprehensive list of changes. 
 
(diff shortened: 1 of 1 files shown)


AaronZyLee (commented):
Could you also add the description about the the default value of FF which is set `true` for new project and `false` for old ones? Otherwise LGTM 💯

FeatureFlags.md
```diff
@@ -0,0 +1,25 @@
...
+- `graphql-docs-generator` - generates GraphQL operations from the given GraphQL schema. Accessed via CLI command `amplify codegen statements` and `amplify-codegen`
+- `graphql-types-generator` - generates type definitions in given target language for the types defined in the GraphQL schema. Accessed via CLI command `amplify codegen types` and `amplify-codegen`
+
+are transferred to this repository from the [Amplify CLI repository](https://github.com/aws-amplify/amplify-cli/tree/master/packages). 
+
+In order to shield the existing customers from any complications as a result of this transfer, we provide the following Feature Flags that enable the customers to switch to the migrated package implementations at their own pace before the deprecation date:
+
+1. **useAppsyncModelgenPlugin**: When set to `true`, the `appsync-modelgen-plugin` implementation in this repository will be used for models generation. When set to `false`, the `amplify-codegen-appsync-model-plugin` implementation from [CLI repo](https://github.com/aws-amplify/amplify-cli/tree/master/packages/amplify-codegen-appsync-model-plugin) will be used for models generation.
```
AaronZyLee, line 10: I find in my `amplify/cli.json` that they are using all lowercases instead of camelCases. Can you double check on your side?
```JSON
{
  "features": {
    "graphqltransformer": {
      "addmissingownerfields": true,
      "validatetypenamereservedwords": true,
      "useexperimentalpipelinedtransformer": false,
      "enableiterativegsiupdates": false,
      "secondarykeyasgsi": true
    },
    "frontend-ios": {
      "enablexcodeintegration": true
    },
    "auth": {
      "enablecaseinsensitivity": true
    },
    "codegen": {
      "useappsyncmodelgenplugin": true,
      "usedocsgeneratorplugin": true,
      "usetypesgeneratorplugin": true
    },
    "appsync": {
      "generategraphqlpermissions": true
    }
  }
}
```

phani-srikar: Yes, CLI seems to convert the flags to lower cases. But [their docs](https://docs.amplify.aws/cli/reference/feature-flags#enableCaseInsensitivity) still maintain the camel case, like how these flags were registered. I believe we should follow their docs convention to avoid confusion. I think irrespective of case, the user can easily figure out what we're referring to here.

AaronZyLee (approved):

PR #99, opened 2021-03-04
Move top level Feature Flag instance calls into the generate Methods

phani-srikar:
_Description of changes:_
The `FeatureFlags.getBoolean` method was used at the top level in types and statements generator commands. This will break because the Feature Flags are not yet initialized. 
Moving these Feature Flag calls next to where they are required i.e inside the `generateTypes` and `generateStatements` methods.

local unit testing


attilah (approved):
LGTM

PR #100, opened 2021-03-04
fix(amplify-codegen): move upstream CLI dependencies as peerDependencies

phani-srikar:
_Description of changes:_
For plugin packages in Node, it is recommended to specify the upstream dependencies on host packages as `peerDependencies` so they do not end up getting their own local versions of those host dependencies in their `node_modules`. 

tested locally by installing beta amplify CLI

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,9 @@
     "rimraf": "^3.0.0",
     "ts-jest": "^25.5.0",
     "ts-node": "^8.10.1",
-    "typescript": "^3.8.3"
+    "typescript": "^3.8.3",
+    "amplify-cli-core": "^1.17.2",
+    "graphql-transformer-core": "^6.26.2"
   },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -83,7 +83,9 @@
     "rimraf": "^3.0.0",
     "ts-jest": "^25.5.0",
     "ts-node": "^8.10.1",
-    "typescript": "^3.8.3"
+    "typescript": "^3.8.3",
+    "amplify-cli-core": "^1.17.2",
```
phani-srikar, line 87: necessary for unit testing as otherwise, `peerDependencies` cannot be included using `require()` construct.

AaronZyLee (approved):
Discuss and test offline. LGTM 🐑

attilah (approved):

PR #114, opened 2021-03-16
feat(modelgen): add timestamp fields createdAt & updatedAt for @model

### LINKED ISSUE #112: Cannot read updatedAt or createdAt in models generated by modelgen
When generating models from a graphql template where you are letting appsync manage the creation and updating of timestamps, `updatedAt` and `createdAt` modelgen does not generate those fields for a customer to be able read from in code.

```
type Clients
  @model(timestamps: { createdAt: "createdOn", updatedAt: "updatedOn" })
  @auth(
    rules: [{ allow: groups, groups: ["companyAdmin"], operations: [create, read, update, delete] }]
  ) {
  id: ID!
  name: String!
  email: String
  tenantId: [ID!]
}
```

In swift this might generate a model like:
```swift
struct Clients {
  let id: UUID
  let name: String
  let email: String
  let tenantId: UUID
}
```
but what we would want is something like:
```swift
struct Clients {
  let id: UUID
  let name: String
  let email: String
  let tenantId: UUID
  let createdAt: Date
  let updatedAt: Date
}
```
so that we can display those properties in the frontend application.

Other issue referenced [here](https://github.com/aws-amplify/amplify-cli/issues/401)

AaronZyLee:
_Issue #, if available:_
fix #112 
_Description of changes:_
Add timestamps fields `createdAt` and `updatedAt` as read-only fields in codegen output for graphql types with `@model`. This is achieved by adding a new metadata `isReadOnly: boolean` in `CodeGenField`. This change is applied to all platforms except for flutter.

The feature will be released behind a feature flag of CLI. Refer https://github.com/aws-amplify/amplify-cli/pull/7273
```TypeScript
export type CodeGenField = TypeInfo & {
  name: string;
  directives: CodeGenDirectives;
  connectionInfo?: CodeGenFieldConnection;
  isReadOnly?: boolean // ---> new metadata
};
```
Related PR from libraries:
Android : https://github.com/aws-amplify/amplify-android/pull/1249
iOS: https://github.com/aws-amplify/amplify-ios/pull/1133

### Implicit schema
Since they are read-only and auto-generated, we suggest to use implicit schema. The timestamp fields are automatically added to modelgen output as an optional type.
1. Simple @model
```GraphQL
type Post @model {
  id: ID!
}
```
2. Use `timestamps` field in @model for implicit schema
You can also use `timestamps` field in @model to change the default timestamp field name. This will add the desired name as optional `AWSDateTime` fields.
```GraphQL
type Post @model(timestamps:{createdAt: "createdOn", updatedAt:"updatedOn"} {
  id: ID!
}
```
### Explicit schema (read-only disabled)
You can also include `createdAt`, `updatedAt`  in the schema field explicitly. However, the timestamp field will be **no longer read-only**. 
```GraphQL
type Post @model {
  id: ID!
  createdAt: AWSDateTime
  updatedAt: AWSDateTime
}
```
This behavior conforms to the fact that graphql transformer will include timestamp fields in input variables in explict schema. 

`yarn test`

### DIFF
--- a/packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-java-visitor.test.ts.snap
+++ b/packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-java-visitor.test.ts.snap
@@ -34,8 +34,8 @@ public final class customClaim implements Model {
   private final @ModelField(targetType=\\"ID\\", isRequired = true) String id;
   private final @ModelField(targetType=\\"String\\") String name;
   private final @ModelField(targetType=\\"String\\") String bar;
-  private @ModelField(targetType=\\"AWSDateTime\\", isReadOnly = true) Temporal.DateTime createdAt = null;
-  private @ModelField(targetType=\\"AWSDateTime\\", isReadOnly = true) Temporal.DateTime updatedAt = null;
+  private @ModelField(targetType=\\"AWSDateTime\\", isReadOnly = true) Temporal.DateTime createdAt;
+  private @ModelField(targetType=\\"AWSDateTime\\", isReadOnly = true) Temporal.DateTime updatedAt;
   public String getId() {
       return id;
   }
@@ -250,8 +250,8 @@ public final class customClaim implements Model {
   private final @ModelField(targetType=\\"ID\\", isRequired = true) String id;
   private final @ModelField(targetType=\\"String\\") String name;
   private final @ModelField(targetType=\\"String\\") String bar;
-  private @ModelField(targetType=\\"AWSDateTime\\", isReadOnly = true) Temporal.DateTime createdAt = null;
-  private @ModelField(targetType=\\"AWSDateTime\\", isReadOnly = true) Temporal.DateTime updatedAt = null;
+  private @ModelField(targetType=\\"AWSDateTime\\", isReadOnly = true) Temporal.DateTime createdAt;
+  private @ModelField(targetType=\\"AWSDateTime\\", isReadOnly = true) Temporal.DateTime updatedAt;
   public String getId() {
       return id;
   }
@@ -471,8 +471,8 @@ public final class Employee implements Model {
   private final @ModelField(targetType=\\"String\\", authRules = {
     @AuthRule(allow = AuthStrategy.OWNER, ownerField = \\"owner\\", identityClaim = \\"cognito:username\\", operations = { ModelOperation.CREATE, ModelOperation.UPDATE, ModelOperation.DELETE, ModelOperation.READ })
   }) String ssn;
-  private @ModelField(targetType=\\"AWSDateTime\\", isReadOnly = true) Temporal.DateTime createdAt = null;
-  private @ModelField(targetType=\\"AWSDateTime\\", isReadOnly = true) Temporal.DateTime updatedAt = null;
+  private @ModelField(targetType=\\"AWSDateTime\\", isReadOnly = true) Temporal.DateTime createdAt;
+  private @ModelField(targetType=\\"AWSDateTime\\", isReadOnly = true) Temporal.DateTime updatedAt;
   public String getId() {
       return id;
   }
@@ -722,8 +722,8 @@ public final class dynamicGroups implements Model {
   private final @ModelField(targetType=\\"ID\\", isRequired = true) String id;
   private final @ModelField(targetType=\\"String\\") String name;
   private final @ModelField(targetType=\\"String\\") String bar;
-  private @ModelField(targetType=\\"AWSDateTime\\", isReadOnly = true) Temporal.DateTime createdAt = null;
-  private @ModelField(targetType=\\"AWSDateTime\\", isReadOnly = true) Temporal.DateTime updatedAt = null;
+  private @ModelField(targetType=\\"AWSDateTime\\", isReadOnly = true) Temporal.DateTime createdAt;
+  private @ModelField(targetType=\\"AWSDateTime\\", isReadOnly = true) Temporal.DateTime updatedAt;
   public String getId() {
       return id;
   }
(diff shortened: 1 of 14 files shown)


packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-swift-visitor.test.ts
```diff
@@ -132,11 +142,17 @@ describe('AppSyncSwiftVisitor', () => {
       public struct snake_case: Model {
         public let id: String
         public var name: String?
+        public var createdAt: Temporal.DateTime?
+        public var updatedAt: Temporal.DateTime?
         
         public init(id: String = UUID().uuidString,
-            name: String? = nil) {
+            name: String? = nil,
+            createdAt: Temporal.DateTime? = nil,
```
kneekey23, line 150: should these even be in the constructor? since they are managed by apppsync I shouldnt be able to set them, only read them?

kneekey23:
overall this LGTM my only concern is that these fields shouldn't be in the constructor? my understanding is that when you use this directive, app sync creates and updates these fields for you and we just want them to be generated on the front end to read from, not set. @undefobj or @renebrandel  can you provide some color here possibly?

undefobj:
> overall this LGTM my only concern is that these fields shouldn't be in the constructor? my understanding is that when you use this directive, app sync creates and updates these fields for you and we just want them to be generated on the front end to read from, not set. @undefobj or @renebrandel  can you provide some color here possibly?

Correct they're set in the GraphQL resolver code not in the client mutations code

kneekey23:
@AaronZyLee per the above confirmation, please remove the generation of these fields in the constructors for each platform, they should be props on the structs or class but not in the constructor.

AaronZyLee:
@kneekey23 @undefobj I agree partially on removing the timestamp fields in constructor. IMO it is dependent on the schema definition. I read the [official doc](https://docs.amplify.aws/cli/graphql-transformer/model#usage) and previous PR https://github.com/aws-amplify/amplify-cli/pull/4149 and find out that there is a back door for users to manipulate the timestamp value if they include them explicitly in the schema. I would suggest that to remove the fields in implicit cases while keep them in explicit cases.

kneekey23:
I agree with that @AaronZyLee , I think that is what I meant actually. In the case of it being implicit, it should not be in the constructor.

packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-java-visitor.test.ts.snap
```diff
@@ -3691,6 +3979,8 @@ public final class Todo implements Model {
   private final @ModelField(targetType=\\"Int\\", isRequired = true) Integer version;
   private final @ModelField(targetType=\\"Float\\") Double value;
   private final @ModelField(targetType=\\"task\\") @HasMany(associatedWith = \\"todo\\", type = task.class) List<task> tasks = null;
+  private @ModelField(targetType=\\"AWSDateTime\\", isReadOnly = true) Temporal.DateTime createdAt = null;
```
richardmcclellan, line 3982: Why do these get set to `null` ?

AaronZyLee: That is a mistake. I will remove them.

diegocstn (commented):
Please refer to this PR https://github.com/aws-amplify/amplify-ios/pull/1133#issue-606644621 for the iOS implementation.
The new fields won't be part of the public initializer but needs to parameters of the internal initializer for encode/decode.

packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-swift-visitor.test.ts
```diff
@@ -41,13 +41,27 @@ describe('AppSyncSwiftVisitor', () => {
...
         public var name: String?
         public var bar: String?
+        public var createdAt: Temporal.DateTime?
+        public var updatedAt: Temporal.DateTime?
         
-        public init(id: String = UUID().uuidString,
-            name: String? = nil,
+        public init(name: String? = nil,
             bar: String? = nil) {
+          self.init(,
```
diegocstn, line 49: here and throughout the generated public initializers, there's an extraneous comma :)

packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-swift-visitor.test.ts
```diff
@@ -132,11 +150,23 @@ describe('AppSyncSwiftVisitor', () => {
       public struct snake_case: Model {
         public let id: String
         public var name: String?
+        public var createdAt: Temporal.DateTime?
+        public var updatedAt: Temporal.DateTime?
         
-        public init(id: String = UUID().uuidString,
-            name: String? = nil) {
+        public init(name: String? = nil) {
+          self.init(,
```
diegocstn, line 157: extraneous comma

diegocstn (approved):

diegocstn (approved):

packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
```diff
@@ -498,6 +507,41 @@ export class AppSyncModelVisitor<
...
+    const updatedAtField: CodeGenField = {
+      name: timestamps?.updatedAt || DEFAULT_UPDATED_TIME,
+      directives: [],
+      type: 'AWSDateTime',
+      isList: false,
+      isNullable: true,
+      isReadOnly: true,
+    };
+    addFieldToModel(model, createdAtField);
+    //updated field will always be override
```
richardmcclellan, line 540: Why is the updated field always overridden?

AaronZyLee: I just tested the graphql transformer by the latest version of CLI and found out that updated field had the same behavior as created field, which means they will both be added to the input variables if defined explicitly. I remove this code in the new commit to aline with CLI.

richardmcclellan (approved):
:shipit:

iartemiev (approved):
Approving the JS/TS changes

djsjr:
Any reason this wasn't done with flutter? Currently need access to createdAt

PR #116, opened 2021-03-17
feat(amplify-codegen): Add Amplify CLI version as a comment for the Modelgen output files

### LINKED ISSUE #90: Add CLI version as a comment in the codegen models
Currently we do not have a mechanism to know if the codegen models are from a particular CLI version or not. This makes it difficult to pin point to an issue. We need to make extra communication with customer to understand the version used. 

Purposed solution:

```swift
// generated by amplify cli version: 2.22.1
// swiftlint:disable all
import Amplify
import Foundation

public struct Blog: Model {
   ...
  }
}
```

In the example above see the line `generated by amplify cli version: 2.22.1`.

phani-srikar:
_Issue #, if available:
fix #90

_Description of changes:_
These changes add the Amplify CLI and Codegen versions used to generate the Datastore Model files as a comment. 
```
// Generated using amplify-cli-version: <version>, amplify-codegen-version: <version>
```
Since all the currently supported language targets (Swift, JS, TS, Java and Dart) use `//` for in-line comments, we did not have to customize the behavior for any specific visitor and handled the version generation in the top level `amplify-codegen` plugin package. 

Most popular programming languages support `//` for in-line comments. [Refer for overview](https://en.wikipedia.org/wiki/Comparison_of_programming_languages_(syntax)#Inline_comments).
The implementation might evolve in the future as we support languages which do not use `//` for in-line comments. 

This PR also comments out a test suite for `cleanGeneratedModels` flag which is not relevant for now. 
Added relevant unit tests.
Tested manually using sample apps. 

### DIFF
--- a/packages/amplify-codegen/src/commands/models.js
+++ b/packages/amplify-codegen/src/commands/models.js
@@ -58,7 +58,7 @@ async function generateModels(context) {
     schema,
     config: {
       target: platformToLanguageMap[projectConfig.frontend] || projectConfig.frontend,
-      directives: directiveDefinitions,
+      directives: directiveDefinitions
     },
   });
 
@@ -87,7 +87,8 @@ async function generateModels(context) {
   appsyncLocalConfig.forEach((cfg, idx) => {
     const outPutPath = cfg.filename;
     fs.ensureFileSync(outPutPath);
-    fs.writeFileSync(outPutPath, generatedCode[idx]);
+    const contentToWrite = [getAmplifyCLIVersionComment(context), generatedCode[idx]].join('\n\n');
+    fs.writeFileSync(outPutPath, contentToWrite);
   });
 
   generateEslintIgnore(context);
@@ -143,6 +144,13 @@ function getModelOutputPath(context) {
     default:
       return '.';
   }
+}
+
+function getAmplifyCLIVersionComment(context) {
+  if (context.usageData == null || !context.usageData.version) {
+    return null
+  }
+  return '// Generated using Amplify CLI version: ' + context.usageData.version;
 }
 
 function generateEslintIgnore(context) {
--- a/packages/amplify-codegen/src/commands/models.js
+++ b/packages/amplify-codegen/src/commands/models.js
@@ -150,7 +150,7 @@ function getAmplifyCLIVersionComment(context) {
   if (context.usageData == null || !context.usageData.version) {
     return null
   }
-  return '// Generated using Amplify CLI version: ' + context.usageData.version;
+  return 'Generated using amplify-cli-version: ' + context.usageData.version;
 }
 
 function generateEslintIgnore(context) {
--- a/packages/amplify-codegen/src/commands/models.js
+++ b/packages/amplify-codegen/src/commands/models.js
@@ -78,18 +78,10 @@ async function generateModels(context) {
 
   const generatedCode = await Promise.all(codeGenPromises);
 
-  // clean the output directory before re-generating models
-  // const cleanOutputPath = FeatureFlags.getBoolean('codegen.cleanGeneratedModelsDirectory');
-  // if (cleanOutputPath) {
-  //   await fs.emptyDir(outputPath);
-  // }
-
   appsyncLocalConfig.forEach((cfg, idx) => {
     const outPutPath = cfg.filename;
     fs.ensureFileSync(outPutPath);
-    const contentsToWrite = [getVersionsMetadataComment(context), generatedCode[idx]].filter(function (content) {
-      return content != null;
-    });
+    const contentsToWrite = [getVersionsMetadataComment(context), generatedCode[idx]].filter(content => content);
     const contentToWrite = contentsToWrite.join('\n\n');
     fs.writeFileSync(outPutPath, contentToWrite);
   });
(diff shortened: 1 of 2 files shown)


packages/amplify-codegen/src/commands/models.js
```diff
@@ -145,6 +146,13 @@ function getModelOutputPath(context) {
   }
 }
 
+function getAmplifyCLIVersionComment(context) {
+  if (context.usageData == null || !context.usageData.version) {
+    return null
+  }
+  return '// Generated using Amplify CLI version: ' + context.usageData.version;
```
diegocstn, line 153: nit: in case we'll ever need/want to use this info programmatically, may be worth using something that is easier to parse like `// amplify-cli-version: X.Y.Z` (similar to what [SwiftPM](https://developer.apple.com/documentation/swift_packages/package) currently does)?

packages/amplify-codegen/tests/commands/models.test.js
```diff
@@ -19,10 +19,13 @@ const MOCK_CONTEXT = {
...
   },
+  usageData: {
+    version: '0.0.0'
+  }
 };
 const OUTPUT_PATHS = {
-    javascript: 'src',
-    android: 'app/src/main/java',
+    javascript: 'src/models',
+    android: 'app/src/main/java/com/amplifyframework/datastore/generated/model',
```
sadiredd-sv, line 28: Why are these paths changed? 
I don't see any change in the model output path fetch in "getModelOutputPath(context)"

packages/amplify-codegen/src/commands/models.js
sadiredd-sv, line 153: nit: is it better to define a const and use it, rather than giving the string here directly?

packages/amplify-codegen/tests/commands/models.test.js
phani-srikar: These point to the actual folders where models are generated.

packages/amplify-codegen/src/commands/models.js
phani-srikar: Changed this to `// Generated using amplify-cli-version: X.Y.Z`. The "Generated" part serves the purpose of warning the user that these files are auto-generated. `amplify-cli-version: X.Y.Z` part of the comment should make parsing easier.

phani-srikar: I am not sure if `const comment = '// Generated using Amplify CLI version: ' + context.usageData.version; return comment` is any different in this case.

packages/amplify-codegen/src/commands/models.js
```diff
@@ -145,6 +146,13 @@ function getModelOutputPath(context) {
   }
 }
 
+function getAmplifyCLIVersionComment(context) {
+  if (context.usageData == null || !context.usageData.version) {
+    return null
+  }
+  return 'Generated using amplify-cli-version: ' + context.usageData.version;
```
AaronZyLee, line 153: CLI uses caret version for amplify-codegen. I think it is better to include this information as well in the comment, which is `Generated using amplify-cli Vxx.xx.xx with amplify-codegen Vxx.xx.xx`

packages/amplify-codegen/src/commands/models.js
sadiredd-sv: What I meant was to declare something like this:
Screen Shot 2021-03-18 at 11 25 50 AM

packages/amplify-codegen/src/commands/models.js
phani-srikar: Added the codegen version. I could only find the information from the context using [this logic](https://github.com/aws-amplify/amplify-codegen/pull/116/files#diff-9cdc392599e83dc816cb33512961b1b923e38faa0692da825ccdaa048081cc6bR159).

packages/amplify-codegen/src/commands/models.js
phani-srikar: This is not exactly a constant string like the `MOCK_GENERATED_CODE`. We are constructing it inside this method and calling it only from a single method (`generateModels`) right now. Adding a global constant might not be a good approach right now.

sadiredd-sv (approved):

sadiredd-sv:
LGTM

packages/amplify-codegen/src/commands/models.js
```diff
@@ -87,7 +87,11 @@ async function generateModels(context) {
   appsyncLocalConfig.forEach((cfg, idx) => {
     const outPutPath = cfg.filename;
     fs.ensureFileSync(outPutPath);
-    fs.writeFileSync(outPutPath, generatedCode[idx]);
+    const contentsToWrite = [getVersionsMetadataComment(context), generatedCode[idx]].filter(function (content) {
+      return content != null;
+    });
```
AaronZyLee, line 92:
```suggestion
    const contentsToWrite = [getVersionsMetadataComment(context), generatedCode[idx]].filter(content => content);
```
You can use the lambda expression instead.

packages/amplify-codegen/src/commands/models.js
```diff
@@ -145,6 +149,29 @@ function getModelOutputPath(context) {
...
+  const versionMetadata = [];
+  if (context.usageData && context.usageData.version) {
+    versionMetadata.push('amplify-cli-version: ' + context.usageData.version);
+  }
+
+  const codegenPluginsInfo = context?.pluginPlatform?.plugins?.codegen;
+  if (codegenPluginsInfo && codegenPluginsInfo.length > 0) {
+    const amplifyCodegenPluginInfo =  codegenPluginsInfo.filter(function(plugin) {
+      return plugin.packageName == 'amplify-codegen';
+    });
```
AaronZyLee, line 163: The same lambda expression can be used here.

AaronZyLee (commented):
Overall LGTM just make sure you run a test with amplify-dev before the release.

AaronZyLee (approved):

diegocstn (approved):

PR #122, opened 2021-03-24
feat(amplify-codegen): add headless support

### LINKED ISSUE #56: Feature Request: Use codegen in headless mode seperate from push
**Is your feature request related to a problem? Please describe.**
When setting up the build configuration for multiple frontend environments with a shared backend, I have the need to make the build process use ```amplify pull``` to get the most up-to-date metadata to connect to the backend (e.x. the ```aws-exports.js``` file for a React webapp). However, to use any of the GraphQL statements generated by ```amplify codegen```, it seems like I need to store a copy of the statements in each one of the frontend repositories in order to use them. It would be preferable if I could generate the statements during/after pulling to reduce manually duplicating the same files across multiple repositories. 

**Describe the solution you'd like**
I think either one of these proposed solutions would work, though the latter would most likely be easier to implement.
- Give an option when running ```amplify pull``` to generate the GraphQL statements if a GraphQL API is detected in the backend (possibly only on the first pull). Something like
```
The current backend environment has a GraphQL API. Would you like to generate new GraphQl statements? (Y/n)
```
Along with the other prompts which appear when running ```amplify codegen```. For headless mode, it would work similarly to how it does for ```amplify push``` by providing a ```--codegen``` flag to the command. 
- Give ```amplify codegen add``` a headless mode, similar to that suggested in aws-amplify/amplify-cli#2005 and aws-amplify/amplify-cli#852. This would allow for automated statement generation if I use ```amplify pull``` to get the entire backend definition, which would include the schema. Alternatively, I could just include the schema file in each repository, which would at least minimize the number of files that need to be manually copied and maintained across each one. 

**Describe alternatives you've considered**
- Using ```amplify pull``` to get the entire backend definition and then using ```amplify push --codegen``` in headless mode. This doesn't work in this case since just pulling the backend means there are no changes and thus ```amplify push``` doesn't run. Even if it did, it wouldn't be ideal to allow the build process to potentially change the backend since it doesn't need to do so.

### LINKED ISSUE #60: Headless support for amplify add codegen
**Is your feature request related to a problem? Please describe.**
I am trying to add multi env support for externally linked APIs but this is currently not possible.

**Describe the solution you'd like**
Headless support for amplify add codegen that lets you pass the codegen config as an argument.

```
CONFIG="{\
\"generateCode\":true,\
\"codeLanguage\":\"typescript\",\
\"fileNamePattern\":\"src/graphql/**/*.ts\",\
\"generatedFileName\":\"src/API.ts\",\
\"generateDocs\":true\
}"

amplify add codegen --apiId xxxxxxxxxxxxxxxxx --config $CONFIG
```

**Describe alternatives you've considered**
Add multi env support for externally linked APIs that allows you to link each environment to a specific AppSync ID.

**Additional context**

AaronZyLee:
_Issue #, if available:_
fix #56 
fix #60
_Description of changes:_
Add headless support for `amplify codegen add`. The change follows the rules of using headless mode in amplify plugins.
`yarn test`


packages/amplify-codegen/src/amplify-plugin-index.js
```diff
@@ -14,7 +16,25 @@ async function handleAmplifyEvent(context, args) {
...
+ */
+async function executeAmplifyHeadlessCommand(context, headlessPayload) {
+  switch (context.input.command) {
+    case 'add':
+      await context.amplify.constructExeInfo(context);
+      context.exeInfo.inputParams[constants.Label] = JSON.parse(headlessPayload);
+      await add(context);
+      break;
+    default:
+      context.print.error(`Headless mode for ${context.input.command} codegen is not implemented yet`);
```
hirenumradia, line 33: The user may pass in a command that is erroneous, should the command be checked to see if it one that Amplify codegen supports? If it is, then it returns the error `Headless mode for ${context.input.command} codegen is not implemented yet ` and if not it says something like `${context.input.command} is not a valid codegen command`

AaronZyLee: Thanks for the comment. `amplify-codegen` is one of the amplify plugins and command validation is conducted in the upper level of `amplify-cli`. You can find the source code here https://github.com/aws-amplify/amplify-cli/blob/99cb0133f38e0579869c71a88213eecb43bb7a8d/packages/amplify-cli/src/input-manager.ts#L96

hirenumradia: Ahh thanks. Makes total sense.

joaopegoraro:
Is there a reason this was closed? I was looking foward to the headless support

PR #142, opened 2021-04-05
ModelGen(Swift, Java): Handle nullability of lists transparently based on the GraphQL schema specification

### LINKED ISSUE #104: Handling of nullable lists and list components in Amplify Datastore
Consider the following GraphQL schema:
```
type ListContainer
@model
{
  id: ID!
  name: String
  list: [String]
  requiredList: [String]!
  requiredListOfRequired: [String!]!
  listOfRequired: [String!]
}
```
GraphQL server allows nullable values for lists and list components [Reference](https://graphql.org/learn/schema/#lists-and-non-null).

`amplify codegen models` for an iOS project generates:
```swift
// swiftlint:disable all
import Amplify
import Foundation

public struct ListContainer: Model {
  public let id: String
  public var name: String?
  public var list: [String]?
  public var requiredList: [String]?
  public var requiredListOfRequired: [String]
  public var listOfRequired: [String]
  
  public init(id: String = UUID().uuidString,
      name: String? = nil,
      list: [String]? = [],
      requiredList: [String]? = [],
      requiredListOfRequired: [String] = [],
      listOfRequired: [String] = []) {
      self.id = id
      self.name = name
      self.list = list
      self.requiredList = requiredList
      self.requiredListOfRequired = requiredListOfRequired
      self.listOfRequired = listOfRequired
  }
}

// swiftlint:disable all
import Amplify
import Foundation

extension ListContainer {
  // MARK: - CodingKeys 
   public enum CodingKeys: String, ModelKey {
    case id
    case name
    case list
    case requiredList
    case requiredListOfRequired
    case listOfRequired
  }
  
  public static let keys = CodingKeys.self
  //  MARK: - ModelSchema 
  
  public static let schema = defineSchema { model in
    let listContainer = ListContainer.keys
    
    model.pluralName = "ListContainers"
    
    model.fields(
      .id(),
      .field(listContainer.name, is: .optional, ofType: .string),
      .field(listContainer.list, is: .optional, ofType: .embeddedCollection(of: String.self)),
      .field(listContainer.requiredList, is: .optional, ofType: .embeddedCollection(of: String.self)),
      .field(listContainer.requiredListOfRequired, is: .required, ofType: .embeddedCollection(of: String.self)),
      .field(listContainer.listOfRequired, is: .required, ofType: .embeddedCollection(of: String.self))
    )
    }
}
```
What should it have generated instead?
```swift
// swiftlint:disable all
import Amplify
import Foundation

public struct ListContainer: Model {
  public let id: String
  public var name: String?
  public var list: [String?]?
  public var requiredList: [String?]
  public var requiredListOfRequired: [String]
  public var listOfRequired: [String]?

  public init(id: String = UUID().uuidString,
      name: String? = nil,
      list: [String?]? = nil,
      requiredList: [String?] = [],
      requiredListOfRequired: [String] = [],
      listOfRequired: [String]? = nil) {
      self.id = id
      self.name = name
      self.list = list
      self.requiredList = requiredList
      self.requiredListOfRequired = requiredListOfRequired
      self.listOfRequired = listOfRequired
  }
}

// swiftlint:disable all
import Amplify
import Foundation

extension ListContainer {
  // MARK: - CodingKeys 
   public enum CodingKeys: String, ModelKey {
    case id
    case name
    case list
    case requiredList
    case requiredListOfRequired
    case listOfRequired
  }

  public static let keys = CodingKeys.self
  //  MARK: - ModelSchema 

  public static let schema = defineSchema { model in
    let listContainer = ListContainer.keys

    model.pluralName = \\"ListContainers\\"

    model.fields(
      .id(),
      .field(listContainer.name, is: .optional, ofType: .string),
      .field(listContainer.list, is: .optional, ofType: .embeddedCollection(of: String.self)),
      .field(listContainer.requiredList, is: .required, ofType: .embeddedCollection(of: String.self)),
      .field(listContainer.requiredListOfRequired, is: .required, ofType: .embeddedCollection(of: String.self)),
      .field(listContainer.listOfRequired, is: .optional, ofType: .embeddedCollection(of: String.self))
    )
    }
}
```
As you see, the Swift types generated for GraphQL lists do not respect the nullability specified in the user's GraphQL schema, leaving him unable to store `null` values for certain fields. The defaults for nullable fields are not `nil` but initialized empty lists which breaks the implementation when the user wants to check for `null`.

phani-srikar:
Issue #, if available:
fix #104

Description of changes:
This PR fixes the bug in handling nullable lists and list components for `iOS` and `Android` platforms. 

There could be various combinations of optional lists in swift like `[String?]?, [String]?, [String?] and [String]`. These have to be generated correctly based on their graphQL definitions `[String], [String!], [String]! and [String!]!`.

Android DataStore on the other hand, does not enforce any checks for nullability of list components. So, a GraphQL field definition like `[String!]` will not be caught during runtime but throws the exception returned from Appsync server when a `null` value is present in the list. This PR includes fix to a bug related to handling the `required` list fields. 

This change will be released behind a Feature Flag. 

How are these changes tested:
Unit test that covers all combinations of nullability in a list and verifies appropriate swift/java type is generated.

### DIFF
--- a/packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-swift-visitor.test.ts.snap
+++ b/packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-swift-visitor.test.ts.snap
@@ -41,12 +41,12 @@ public struct ListContainer: Model {
   }
   internal init(id: String = UUID().uuidString,
       name: String? = nil,
-      list: [Int?] = [],
+      list: [Int?]? = nil,
       requiredList: [String?] = [],
       requiredListOfRequired: [StatusEnum] = [],
-      listOfRequired: [Bool] = [],
+      listOfRequired: [Bool]? = nil,
       requiredListOfRequiredDates: [Temporal.Date] = [],
-      listOfRequiredFloats: [Double] = [],
+      listOfRequiredFloats: [Double]? = nil,
       requiredListOfCustomTypes: [CustomType?] = [],
       createdAt: Temporal.DateTime? = nil,
       updatedAt: Temporal.DateTime? = nil) {
@@ -125,7 +125,7 @@ public struct CustomType: Embeddable {
 }"
 `;
 
-exports[`AppSyncSwiftVisitor Should render lists with HAS_MANY connection according to the spec 1`] = `
+exports[`AppSyncSwiftVisitor Should render lists with HAS_MANY connection as optional fields 1`] = `
 "// swiftlint:disable all
 import Amplify
 import Foundation
@@ -133,34 +133,49 @@ import Foundation
 public struct Todo: Model {
   public let id: String
   public var title: String
-  public var tasks: List<task?>
+  public var requiredListOfTasks: List<task>?
+  public var listOfRequiredTasks: List<task>?
+  public var listOfTasks: List<task>?
+  public var requiredListOfRequiredTasks: List<task>?
   public var createdAt: Temporal.DateTime?
   public var updatedAt: Temporal.DateTime?
   
   public init(id: String = UUID().uuidString,
       title: String,
-      tasks: List<task?> = []) {
+      requiredListOfTasks: List<task>? = [],
+      listOfRequiredTasks: List<task> = [],
+      listOfTasks: List<task>? = [],
+      requiredListOfRequiredTasks: List<task> = []) {
     self.init(id: id,
       title: title,
-      tasks: tasks,
+      requiredListOfTasks: requiredListOfTasks,
+      listOfRequiredTasks: listOfRequiredTasks,
+      listOfTasks: listOfTasks,
+      requiredListOfRequiredTasks: requiredListOfRequiredTasks,
       createdAt: nil,
       updatedAt: nil)
   }
   internal init(id: String = UUID().uuidString,
       title: String,
-      tasks: List<task?> = [],
+      requiredListOfTasks: List<task>? = [],
+      listOfRequiredTasks: List<task> = [],
+      listOfTasks: List<task>? = [],
+      requiredListOfRequiredTasks: List<task> = [],
       createdAt: Temporal.DateTime? = nil,
       updatedAt: Temporal.DateTime? = nil) {
       self.id = id
       self.title = title
-      self.tasks = tasks
+      self.requiredListOfTasks = requiredListOfTasks
+      self.listOfRequiredTasks = listOfRequiredTasks
+      self.listOfTasks = listOfTasks
+      self.requiredListOfRequiredTasks = requiredListOfRequiredTasks
       self.createdAt = createdAt
       self.updatedAt = updatedAt
   }
 }"
 `;
 
-exports[`AppSyncSwiftVisitor Should render lists with HAS_MANY connection according to the spec 2`] = `
+exports[`AppSyncSwiftVisitor Should render lists with HAS_MANY connection as optional fields 2`] = `
 "// swiftlint:disable all
 import Amplify
 import Foundation
@@ -170,7 +185,10 @@ extension Todo {
    public enum CodingKeys: String, ModelKey {
     case id
     case title
-    case tasks
+    case requiredListOfTasks
+    case listOfRequiredTasks
+    case listOfTasks
+    case requiredListOfRequiredTasks
     case createdAt
     case updatedAt
   }
@@ -186,7 +204,10 @@ extension Todo {
     model.fields(
       .id(),
       .field(todo.title, is: .required, ofType: .string),
-      .hasMany(todo.tasks, is: .required, ofType: task.self, associatedWith: task.keys.todo),
+      .hasMany(todo.requiredListOfTasks, is: .optional, ofType: task.self, associatedWith: task.keys.todo),
+      .hasMany(todo.listOfRequiredTasks, is: .optional, ofType: task.self, associatedWith: task.keys.todo),
+      .hasMany(todo.listOfTasks, is: .optional, ofType: task.self, associatedWith: task.keys.todo),
+      .hasMany(todo.requiredListOfRequiredTasks, is: .optional, ofType: task.self, associatedWith: task.keys.todo),
       .field(todo.createdAt, is: .optional, isReadOnly: true, ofType: .dateTime),
       .field(todo.updatedAt, is: .optional, isReadOnly: true, ofType: .dateTime)
     )
@@ -194,7 +215,7 @@ extension Todo {
 }"
 `;
 
-exports[`AppSyncSwiftVisitor Should render lists with HAS_MANY connection according to the spec 3`] = `
+exports[`AppSyncSwiftVisitor Should render lists with HAS_MANY connection as optional fields 3`] = `
 "// swiftlint:disable all
 import Amplify
 import Foundation
@@ -229,7 +250,7 @@ public struct task: Model {
 }"
 `;
 
-exports[`AppSyncSwiftVisitor Should render lists with HAS_MANY connection according to the spec 4`] = `
+exports[`AppSyncSwiftVisitor Should render lists with HAS_MANY connection as optional fields 4`] = `
 "// swiftlint:disable all
 import Amplify
 import Foundation
--- a/packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-swift-visitor.test.ts.snap
+++ b/packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-swift-visitor.test.ts.snap
@@ -1,0 +1,66 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`AppSyncSwiftVisitor Should handle nullability of lists appropriately 1`] = `
+"// swiftlint:disable all
+import Amplify
+import Foundation
+
+public struct ListContainer: Model {
+  public let id: String
+  public var name: String?
+  public var list: [String?]?
+  public var requiredList: [String?]
+  public var requiredListOfRequired: [String]
+  public var listOfRequired: [String]?
+  
+  public init(id: String = UUID().uuidString,
+      name: String? = nil,
+      list: [String?]? = [],
+      requiredList: [String?] = [],
+      requiredListOfRequired: [String] = [],
+      listOfRequired: [String]? = []) {
+      self.id = id
+      self.name = name
+      self.list = list
+      self.requiredList = requiredList
+      self.requiredListOfRequired = requiredListOfRequired
+      self.listOfRequired = listOfRequired
+  }
+}"
+`;
+
+exports[`AppSyncSwiftVisitor Should handle nullability of lists appropriately 2`] = `
+"// swiftlint:disable all
+import Amplify
+import Foundation
+
+extension ListContainer {
+  // MARK: - CodingKeys 
+   public enum CodingKeys: String, ModelKey {
+    case id
+    case name
+    case list
+    case requiredList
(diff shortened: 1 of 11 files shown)


richardmcclellan (approved):
LGTM!

packages/appsync-modelgen-plugin/src/visitors/appsync-swift-visitor.ts
```diff
@@ -236,7 +239,11 @@ export class AppSyncSwiftVisitor extends AppSyncModelVisitor {
     const name = `${modelKeysName}.${this.getFieldName(field)}`;
     const typeName = this.getSwiftModelTypeName(field);
     const { connectionInfo } = field;
-    const isRequired = this.isFieldRequired(field) ? '.required' : '.optional';
+    let isOptionalField = field.isList ? field.isListNullable : field.isNullable
+    if (field.connectionInfo && field.connectionInfo.kind === CodeGenConnectionType.HAS_MANY) {
+      isOptionalField = true;
```
AaronZyLee, line 244: The `HAS_MANY` field will always be optional. Is this a related change?

packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-swift-visitor.test.ts
```diff
@@ -275,6 +275,48 @@ describe('AppSyncSwiftVisitor', () => {
...
+    const visitor = getVisitor(schema, 'ListContainer');
+    const generatedCode = visitor.generate();
+    expect(generatedCode).toMatchSnapshot();
+
+    const metadataVisitor = getVisitor(schema, 'ListContainer', CodeGenGenerateEnum.metadata);
+    const generatedMetadata = metadataVisitor.generate();
+    expect(generatedMetadata).toMatchSnapshot();
+
+    const customTypeVisitor = getVisitor(schema, 'CustomType');
+    expect(customTypeVisitor.generate()).toMatchSnapshot();
```
AaronZyLee, line 317: The swift test is using the inline snapshot for now. Although it is better to use a separate file for snapshot, we should keep all of them in one place. I suggest to change all inline snapshots into one file or keep this test as inline temporarily and move them in another PR.

packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-swift-visitor.test.ts.snap
```diff
@@ -0,0 +1,95 @@
...
+  
+  public init(id: String = UUID().uuidString,
+      name: String? = nil,
+      list: [Int?]? = [],
+      requiredList: [String?] = [],
+      requiredListOfRequired: [StatusEnum] = [],
+      listOfRequired: [Bool]? = [],
+      requiredListOfRequiredDates: [Temporal.Date] = [],
+      listOfRequiredFloats: [Double]? = [],
+      requiredListOfCustomTypes: [CustomType?] = []) {
```
AaronZyLee, line 27: For a listNullable field like `[var]` or `[var!]`, should it be initialized as `nil` instead?

lawmicha (approved):
@phani-srikar last time you had provided code generated models for me to use in this test in the library: https://github.com/aws-amplify/amplify-ios/pull/1145/files#diff-532a3db572dc2274087ce086e388eff753b54d3e995fa02a0c344e1097433dd9 as long as the codegen result is the same then LGTM!

packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-swift-visitor.test.ts
phani-srikar: Keeping snapshots separate is preferred over inline snaps. I am not sure why we should keep all inline snaps for now?

packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-swift-visitor.test.ts.snap
phani-srikar: @lawmicha Do you think we should change these defaults to `nil` for optional lists?

packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-swift-visitor.test.ts
AaronZyLee: What I mean is that we should keep the snapshots into one place. I agree it is better to keep them in a separate file.However, you just add the new case in snapshot file but keep the old ones as inline.

packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-swift-visitor.test.ts.snap
lawmicha: are there any default values for other platforms that we have to make a similar decision? Seems like the right thing to do is to set it to the most empty value, which would be nil

lawmicha: but in doing so, we should re-evaluate what that looks like for developers in terms of the upgrade (when the regenerate the files, do they have any code changes to do?). before, they may have been able to do something like we just need to communicate to the developers the change to default values in-addition to the change in nullability of each type
```
let listContainer = ListContainer() // default creates empty lists
listContainer.listofRequiredFloats?.append(1.0) // directly append, without needing to instantiate it
```

phani-srikar: Java does not support default values for method parameters but all the list variables (optional or required) will be `null` if not initialized. In typescript they will be `undefined` if not initialized. I agree that setting the optionals to `nil` by default makes sense for Swift.

lawmicha:
Can we add a link to platform specific guides when customer runs `amplify codegen models` to communicate migration plan if they have upgraded the CLI and are regenerating their models? 
pointing to a platform specific guide here:  https://docs.amplify.aws/guides/q/platform/ios

packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-swift-visitor.test.ts.snap
phani-srikar: Changed the defaults to `nil` for optional types.

packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-swift-visitor.test.ts.snap
```diff
@@ -0,0 +1,264 @@
...
+
+exports[`AppSyncSwiftVisitor Should render lists with HAS_MANY connection according to the spec 1`] = `
+"// swiftlint:disable all
+import Amplify
+import Foundation
+
+public struct Todo: Model {
+  public let id: String
+  public var title: String
+  public var tasks: List<task?>
```
phani-srikar, line 136: pay attention to this type. Since the schema has `tasks: [task]! @connection(name: "TodoTasks")`, the list is required here.

packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-swift-visitor.test.ts.snap
```diff
@@ -0,0 +1,264 @@
...
+  
+  public static let schema = defineSchema { model in
+    let todo = Todo.keys
+    
+    model.pluralName = \\"Todos\\"
+    
+    model.fields(
+      .id(),
+      .field(todo.title, is: .required, ofType: .string),
+      .hasMany(todo.tasks, is: .required, ofType: task.self, associatedWith: task.keys.todo),
```
phani-srikar, line 189: This will be `.required`. Hope this is expected @lawmicha

lawmicha: isn't this PR only applicable to embedded types? why are we changing connection types?

lawmicha: what is the previous `is:` field? `.optional` or `.required`?

lawmicha:
latest generated models do not compile
```swift
// swiftlint:disable all
import Amplify
import Foundation

public struct Blog6: Model {
  public let id: String
  public var name: String
  public var posts: List<Post6>?
  
  public init(id: String = UUID().uuidString,
      name: String,
      posts: List<Post6>? = []) {
    self.init(id: id,
      name: name,
      posts: posts)
  }
  internal init(id: String = UUID().uuidString,
      name: String,
      posts: List<Post6>? = []) {
      self.id = id
      self.name = name
      self.posts = posts
  }
}
```

Error:
`Invalid redeclaration of 'init(id:name:posts:)'`

diegocstn:
> latest generated models do not compile
> 
> ```swift
> // swiftlint:disable all
> import Amplify
> import Foundation
> 
> public struct Blog6: Model {
>   public let id: String
>   public var name: String
>   public var posts: List<Post6>?
>   
>   public init(id: String = UUID().uuidString,
>       name: String,
>       posts: List<Post6>? = []) {
>     self.init(id: id,
>       name: name,
>       posts: posts)
>   }
>   internal init(id: String = UUID().uuidString,
>       name: String,
>       posts: List<Post6>? = []) {
>       self.id = id
>       self.name = name
>       self.posts = posts
>   }
> }
> ```
> 
> Error:
> `Invalid redeclaration of 'init(id:name:posts:)'`

it might be related to the same issue fixed by https://github.com/aws-amplify/amplify-codegen/pull/173

packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-swift-visitor.test.ts.snap
```diff
@@ -0,0 +1,285 @@
...
+      self.requiredList = requiredList
+      self.requiredListOfRequired = requiredListOfRequired
+      self.listOfRequired = listOfRequired
+      self.requiredListOfRequiredDates = requiredListOfRequiredDates
+      self.listOfRequiredFloats = listOfRequiredFloats
+      self.requiredListOfCustomTypes = requiredListOfCustomTypes
+      self.createdAt = createdAt
+      self.updatedAt = updatedAt
+  }
+}"
```
lawmicha, line 65: how come there is an internal init here?

lawmicha: @diegocstn when timestamp is enabled, then there is an internal init? when it is not, then there is no internal init? what does timestamp enablement mean?

lawmicha (approved):
latest generated code looks good! iOS PR for updating to latest models and API/DataStore integ testing is completed here: https://github.com/aws-amplify/amplify-ios/pull/1145

packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-swift-visitor.test.ts.snap
diegocstn: @lawmicha `timestampEnabled` is a feature flag to support `createdAt` and `updatedAt` fields, It's enabled by default on new projects. The internal initializer is only used internally to decode model instances with read-only fields. The public initializer doesn't accept `createdAt` and `updatedAt` parameters

PR #146, opened 2021-04-08
chore: optimize local testing of amplify-dev

AaronZyLee:
_Issue #, if available:_

_Description of changes:_
The changes enable developers to use `amplify-dev` directly for local testing by just one command `yarn setup-dev`. The cli binary in `node_modules` can also be used for local amplify e2e tests.

- Add `@aws-amplify/cli` with caret version(latest) as the dev-dependency.
- Remove additional dependencies(cli-core, graphql-core) since they are already included in cli.
- Add scripts to hoist cli package in `node_modules` so that it can scan the related plugins successfully.
- Add scripts to link `amplify-dev` to yarn global

- `yarn setup-dev` and run `amplify-dev -v`
- Run `amplify-dev codegen [subcommands]` to make sure it runs successfully and reflects local changes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,24 +4,52 @@ Thank you for your interest in contributing to our project! 💛
 
 Whether it's a bug report, new feature, correction, or additional documentation, we greatly value feedback and contributions from our community. Please read through these guidelines carefully before submitting a PR or issue and let us know if it's not up-to-date (or even better, submit a PR with your proposed corrections 😉).
 
-# Development Process
+## Getting Started
 
 Our work is done directly on Github and PR's are sent to the GitHub repo by core team members and contributors. Everyone undergoes the same review process to get their changes into the repo.
 
-# Setting up for local development
+This section should get you running with **Amplify Codegen**.
 
-This section should get you running with **Amplify Codegen**. You will need the latest version of [nodejs](https://nodejs.org/en/) on your system and developing locally also requires `yarn` workspaces. You can install it [here](https://classic.yarnpkg.com/en/docs/install#mac-stable).
+### Setting up for local development
 
-Start by, [Forking](https://help.github.com/en/github/getting-started-with-github/fork-a-repo) the main branch of [amplify-codegen](https://github.com/aws-amplify/amplify-codegen).
+1. You will need the latest version of [nodejs](https://nodejs.org/en/) on your system and developing locally also requires `yarn` workspaces. You can install it [here](https://classic.yarnpkg.com/en/docs/install#mac-stable).
 
-```
+2. Start by [Forking](https://help.github.com/en/github/getting-started-with-github/fork-a-repo) the main branch of [amplify-codegen](https://github.com/aws-amplify/amplify-codegen).
+
+```sh
 $ git clone git@github.com:[username]/amplify-codegen.git
-$ cd amplify-codegen
-
-$ yarn run setup-dev
 ```
 
-# Bugs
+> NOTE: Make sure to always [sync your fork](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/syncing-a-fork) with _main_ branch of amplify-codegen
+
+3. Move into your project folder
+
+```sh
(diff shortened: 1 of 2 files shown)


phani-srikar:
`amplify-dev` is the binary Amplify CLI uses to test their local changes to CLI packages. In our scenario, this binary will be from the latest released CLI package from NPM and our local codegen plugin, which can then be used to test our local codegen changes. For a Amplify CLI developer, this might be a bit confusing if we use the same name to reference both binaries. My suggestion is to either call this out explicitly in the [contribution doc](https://github.com/aws-amplify/amplify-codegen/blob/master/CONTRIBUTING.md#setting-up-for-local-development) that we override the behavior of `amplify-dev` or rename our reference to `amplify-codegen-dev`.

frimfram (approved):
We tested this one and it worked well for dev env.

AaronZyLee:
> `amplify-dev` is the binary Amplify CLI uses to test their local changes to CLI packages. In our scenario, this binary will be from the latest released CLI package from NPM and our local codegen plugin, which can then be used to test our local codegen changes. For a Amplify CLI developer, this might be a bit confusing if we use the same name to reference both binaries. My suggestion is to either call this out explicitly in the [contribution doc](https://github.com/aws-amplify/amplify-codegen/blob/master/CONTRIBUTING.md#setting-up-for-local-development) that we override the behavior of `amplify-dev` or rename our reference to `amplify-codegen-dev`.

I agree to explain it in the document that the behavior will be override. We cannot change the name because of this [code snippet](https://github.com/aws-amplify/amplify-cli/blob/a46f2a32ec9bf9e75684bc93a2e7089ac3fb894d/packages/amplify-cli/src/plugin-helpers/access-plugins-file.ts#L34-L36) unless there is additional change to cli. In addition, there is no impact for cli developers by using `amplify-dev` if they run the command under cli repo because the new link will always override the old one.

CONTRIBUTING.md
```diff
@@ -40,20 +68,20 @@ Guidelines for bug reports:
...
 
 Pull requests are welcome!
 
-You should open an issue to discuss your pull request, unless it's a trivial change. It's best to ensure that your proposed change would be accepted so that you don't waste your own time. If you would like to implement support for a significant feature that is not yet available, please talk to us beforehand to avoid any duplication of effort. 
+You should open an issue to discuss your pull request, unless it's a trivial change. It's best to ensure that your proposed change would be accepted so that you don't waste your own time. If you would like to implement support for a significant feature that is not yet available, please talk to us beforehand to avoid any duplication of effort.
 
 Pull requests should generally be opened against **master**.
 
-Only include ***src*** files in your PR. Don't include any build files i.e. dist/. These will be built upon publish to npm and when a release is created on GitHub.
+Don't include any build files i.e. `dist/`. These will be built upon publish to npm and when a release is created on GitHub.
```
phani-srikar, line 79: we can mention to not include `lib/` files too

CONTRIBUTING.md
```diff
@@ -4,24 +4,52 @@ Thank you for your interest in contributing to our project! 💛
...
+
+### Architecture of the codebase
+
+Amplify Codegen is a monorepo built with [Yarn Workspaces](https://yarnpkg.com/features/workspaces) and [Lerna](https://github.com/lerna/lerna). All packages live within the `packages/` directory in the root. Each category inside packages has its own `src/` and `package.json`.
+
+### Steps towards contribution
+
+- Make changes to required file.
+- Write unit tests
+- Yarn build
```
phani-srikar, line 47: Could be rephrased as: Run `yarn build` to compile your changes

CONTRIBUTING.md
```diff
@@ -4,24 +4,52 @@ Thank you for your interest in contributing to our project! 💛
...
+- Run test suite
+- Test in sample app using [amplify-dev](#tests)
```
phani-srikar, line 49: should the link be `#Tests` ? Also include that link in above point `Run test suite`

phani-srikar (approved):
LGTM

PR #156, opened 2021-04-23
Split e2e tests based on frontend. Add e2e tests for codegen configure and add workflows

phani-srikar:
_Description of changes:_
Includes P0 cases to test the `amplify codegen add` and `amplify codegen configure` workflows. 
The tests for each frontend are split into a separate file. This brings down the time to run the test-suite because `jest` runs each of them concurrently. The number of test files run concurrently depends on the number of available CPU threads (except the main thread).

Each test suite being added is tested locally.


packages/amplify-codegen-e2e-tests/src/__tests__/add-codegen.test.ts
```diff
@@ -0,0 +1,116 @@
...
+        // pre-existing file should still exist
+        expect(existsSync(userSourceCodePath)).toBe(true);
+        // GraphQL statements are generated
+        expect(isNotEmptyDir(path.join(projectRoot, config.graphqlCodegenDir))).toBe(true);
+        // graphql configuration should be added
+        expect(existsSync(path.join(projectRoot, graphqlConfigFile))).toBe(true);
+        // check if the added codegen configuration is valid
+        expect(readFileSync(path.join(projectRoot, graphqlConfigFile)).toString()).toMatchSnapshot();
+      });
+    });
```
AaronZyLee, line 115: These tests are the same as the tests in `graphql-codegen` except for the snapshots. https://github.com/aws-amplify/amplify-codegen/blob/f0e6857d84d2e7fe91969881a9aadaeac5c9e533/packages/amplify-codegen-e2e-tests/src/__tests__/graphql-codegen.test.ts#L41-L51

AaronZyLee (commented):
I would suggest to add codegen add/configure/remove tests into one suite `graphql-codegen`. Some of the tests are duplicated of the original tests. Besides, the provisioning of amplify resources as `amplify init/push/delete` cosumes a lot of time and are the dominant time slot of the total time. It is better to reduce the usage of them by combining some cases into one provision run so that the efficiency is improved.

packages/amplify-codegen-e2e-tests/src/__tests__/add-codegen.test.ts
phani-srikar: Testing various user flows in a single test suite might not be optimal. We can reduce the time taken for running the e2e tests by running each frontend being tested in parallel and also splitting the test suites into a separate job in CircleCI. I will include the circleci changes in a separate PR.

packages/amplify-codegen-e2e-tests/src/__tests__/remove-codegen-js.test.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    beforeEach(async () => {
+        projectRoot = await createNewProjectDir('removeCodegenJS');
+    });
+
+    afterEach(async () => {
+        await deleteAmplifyProject(projectRoot);
+    });
+
+    it(`Do nothing during remove when codegen is not added in JS project`, async () => {
```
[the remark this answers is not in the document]
AaronZyLee: On the other hand, the `codegen remove` command does not have a strong relationship with the frontend type. It does have close relation to whether you add codegen or not.  I would suggest not divide the remove test based on different frontends.

packages/amplify-codegen-e2e-tests/src/codegen-tests-base/test-setup.ts
```diff
@@ -0,0 +1,71 @@
...
+    deleteProjectDir(projectRoot);
+}
+
+export function testSetupBeforeAddCodegen(projectRoot: string, config: AmplifyFrontendConfig): string {
+    // generate pre-existing user file
+    const userSourceCodePath = generateSourceCode(projectRoot, config.srcDir);
+
+    // verify that the pre-existing file exists and codegen output directory is empty
+    expect(existsSync(userSourceCodePath)).toBe(true);
+    expect(isNotEmptyDir(path.join(projectRoot, config.graphqlCodegenDir))).toBe(false);
```
AaronZyLee, line 35: Before adding the codegen, this directory does not even exist rather than `isNotEmpty`.

packages/amplify-codegen-e2e-core/src/utils/graphql-config-helper.ts
```diff
@@ -0,0 +1,51 @@
+import { AmplifyFrontendConfig, AmplifyFrontend } from './frontend-config-helper';
+
+export function constructGraphQLConfig(
```
AaronZyLee, line 3: It looks like the field such as `apiId`, `generatedFileName` is not configurable in this function.

AaronZyLee (commented):
It is good to see the tests are split by the frontend types but we still need to divide each test suites into an individual job on circleCI. 

As for the case itself, I left some comments. Not sure if you run some of the test locally. I assume there should be some failing cases. Once you pass all the tests locally the PR can be merged to see the results from CircleCI.

packages/amplify-codegen-e2e-tests/src/codegen-tests-base/test-setup.ts
phani-srikar: Changed this to test that the directory does not exist

packages/amplify-codegen-e2e-tests/src/__tests__/remove-codegen-js.test.ts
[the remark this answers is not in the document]
phani-srikar: `codegen remove` without add does not do much except throw a useful message to run `codegen add`. So haven't tested this for all frontends. However, the second case that is tested for all frontends is to make sure we do not delete any pre-existing or generated files during `codegen remove`. Anything that says remove/delete is scary to be not tested right now :)

packages/amplify-codegen-e2e-core/src/utils/graphql-config-helper.ts
phani-srikar: Added `apiId`. `generatedFileName` depends on the frontend config being passed. It is set accordingly.

phani-srikar:
> It is good to see the tests are split by the frontend types but we still need to divide each test suites into an individual job on circleCI(out of the scope of current PR).
> 
> As for the case itself, I left some comments. Not sure if you run some of the test locally. I assume there should be some failing cases. Once you pass all the tests locally the PR can be merged to see the results from CircleCI.

Tested these new tests locally and on circleCI. I did observe that `codegen pull` tests were failing yesterday. But that is not likely caused by this change as it does not change that test suite.

AaronZyLee (approved):
Tested pull tests locally and no failure occurred. Merge it to e2e branch to see the performance on circleCI.

PR #173, opened 2021-05-20
fix(modelgen): iOS - skip internal initializer based on feature flag

diegocstn:
#### Description of changes
This PR fixes an issue where a duplicate internal constructor is generated in Swift models.
If the `addTimestampFields` is not set or false, the internal initializer used to deserialize Swift models should not be generated as will result in an `invalid redeclaration` error because will have the exact same set of parameters of the public initializer.

invalid-redeclaration-error

- `yarn test`
- `amplify-dev codegen models` in an iOS project with and without the `addTimestampFields` feature flag

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-codegen/src/commands/models.js
+++ b/packages/amplify-codegen/src/commands/models.js
@@ -55,7 +55,7 @@ async function generateModels(context) {
 
   let isTimestampFieldsAdded = false;
   try {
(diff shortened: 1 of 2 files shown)


AaronZyLee:
Close this in favor of #177

PR #185, opened 2021-06-09
feat(appsync-modelgen-plugin): expose readonly labels to model ts output

nickarocho:
#### Description of changes

This allows us to omit the readonly fields (only if the `config.isTimestampFieldsAdded` flag is enabled in the `cli.json` within a user's project) from being suggested by Intellisense. The associated tests were updated automatically via running `jest --updateSnapshot`.

**_Additionally_**, I've extended the `typescript-declaration-block.ts` file with a new method to dynamically generate `type` blocks which get output in the `index.d.ts` file in user projects. I'm updating the `types.ts` in JS Datastore to reflect this new second parameter passed in to the constructor/copyOf (as seen in the generated snapshots in the test file).

#### Issue #, if available

None

`yarn test`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed](https://github.com/aws-amplify/amplify-codegen/blob/master/CONTRIBUTING.md#tests)

### DIFF
--- a/packages/appsync-modelgen-plugin/src/visitors/appsync-javascript-visitor.ts
+++ b/packages/appsync-modelgen-plugin/src/visitors/appsync-javascript-visitor.ts
@@ -52,6 +52,10 @@ export class AppSyncModelJavascriptVisitor<
         .map(enumObj => this.generateEnumDeclarations(enumObj, true))
         .join('\n\n');
 
+      const modelMetaData = Object.values(this.modelMap)
+        .map(typeObj => this.generateModelMetaData(typeObj))
+        .join('\n\n');
+
       const modelDeclarations = Object.values(this.modelMap)
         .map(typeObj => this.generateModelDeclaration(typeObj, true))
         .join('\n\n');
@@ -60,7 +64,7 @@ export class AppSyncModelJavascriptVisitor<
         .map(typeObj => this.generateModelDeclaration(typeObj, true))
         .join('\n\n');
 
-      return [imports, enumDeclarations, nonModelDeclarations, modelDeclarations].join('\n\n');
+      return [imports, enumDeclarations, nonModelDeclarations, modelMetaData, modelDeclarations].join('\n\n');
     } else {
       const imports = this.generateImportsJavaScriptImplementation();
       const enumDeclarations = Object.values(this.enumMap)
--- a/packages/appsync-modelgen-plugin/src/visitors/appsync-javascript-visitor.ts
+++ b/packages/appsync-modelgen-plugin/src/visitors/appsync-javascript-visitor.ts
@@ -52,10 +52,6 @@ export class AppSyncModelJavascriptVisitor<
         .map(enumObj => this.generateEnumDeclarations(enumObj, true))
         .join('\n\n');
 
-      const modelMetaData = Object.values(this.modelMap)
-        .map(typeObj => this.generateModelMetaData(typeObj))
-        .join('\n\n');
-
       const modelDeclarations = Object.values(this.modelMap)
         .map(typeObj => this.generateModelDeclaration(typeObj, true))
         .join('\n\n');
@@ -64,7 +60,15 @@ export class AppSyncModelJavascriptVisitor<
         .map(typeObj => this.generateModelDeclaration(typeObj, true))
         .join('\n\n');
 
-      return [imports, enumDeclarations, nonModelDeclarations, modelMetaData, modelDeclarations].join('\n\n');
+      if (!this.config.isTimestampFieldsAdded) {
+        return [imports, enumDeclarations, nonModelDeclarations, modelDeclarations].join('\n\n');
+      } else {
+        const modelMetaData = Object.values(this.modelMap)
+          .map(typeObj => this.generateModelMetaData(typeObj))
+          .join('\n\n');
+
+        return [imports, enumDeclarations, nonModelDeclarations, modelMetaData, modelDeclarations].join('\n\n');
+      }
     } else {
       const imports = this.generateImportsJavaScriptImplementation();
       const enumDeclarations = Object.values(this.enumMap)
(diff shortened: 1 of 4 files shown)


AaronZyLee (commented):
The change looks good. Just have a question for the old users. Will these change break their applications? Since read-only fields are added behind the CLI's feature flag, the old projects will not have timestamp fields added unless they manually switch the feature flag. If it breaks the old projects, we need to consider the implementation based on the feature flag value. 

The code snippet is as following FYI. https://github.com/aws-amplify/amplify-codegen/blob/abb966c647b08537564f85314a4266a5cbe33ffd/packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts#L523-L525

packages/appsync-modelgen-plugin/src/visitors/appsync-typescript-visitor.ts
```diff
@@ -117,7 +125,7 @@ export class AppSyncModelTypeScriptVisitor<
           },
           {
             name: 'mutator',
-            type: `(draft: MutableModel<${modelName}>) => MutableModel<${modelName}> | void`,
+            type: `(draft: MutableModel<${modelName}${readOnlyTypeLabels}>) => MutableModel<${modelName}${readOnlyTypeLabels}> | void`,
```
frimfram, line 128: These changes would need to be isolated under a feature flag.  (I think Aaron pointed out feature flag as well)

frimfram (changes requested):

packages/appsync-modelgen-plugin/src/visitors/appsync-typescript-visitor.ts
nickarocho: @frimfram I've added a check for the feature flag, please verify. Thank you!

AaronZyLee (approved):
LGTM :octocat:

nickarocho:
Hi @AaronZyLee and @frimfram, I've made a few adjustments to this PR based on Typescript requirements in the Datastore JS library. I've added a new method to extend the class that generates classes to now generate types. Could you please give it another look when you have a chance and let me know if you have any questions or concerns? Thanks!

packages/appsync-modelgen-plugin/src/visitors/appsync-javascript-visitor.ts
```diff
@@ -52,6 +52,10 @@ export class AppSyncModelJavascriptVisitor<
         .map(enumObj => this.generateEnumDeclarations(enumObj, true))
         .join('\n\n');
 
+      const modelMetaData = Object.values(this.modelMap)
+        .map(typeObj => this.generateModelMetaData(typeObj))
+        .join('\n\n');
+
```
frimfram, line 58: These types of changes need to go under the feature flag.

packages/appsync-modelgen-plugin/src/visitors/appsync-javascript-visitor.ts
```diff
@@ -60,7 +64,7 @@ export class AppSyncModelJavascriptVisitor<
         .map(typeObj => this.generateModelDeclaration(typeObj, true))
         .join('\n\n');
 
-      return [imports, enumDeclarations, nonModelDeclarations, modelDeclarations].join('\n\n');
+      return [imports, enumDeclarations, nonModelDeclarations, modelMetaData, modelDeclarations].join('\n\n');
```
frimfram, line 67: Again, inserting a new field like modelMetaData in the middle of an array is not ideal for backward compatibility.

frimfram (changes requested):
Overall, I have concerns about the backward compatibility.  
Suggestion is that instead of changing the behavior of the existing function, maybe create a new function that calls the old one with additional new behaviors.  
For example, instead of changing "generate" function, leave the old generate function as is and create a new one "generateWithTypes" that calls the original generate and then create more objects to be returned along with the results from the original.

AaronZyLee (changes requested):
I would suggest a change of the function `getVisitor` in the js visitor test file. Right now the `isTimestampFieldAdded` is always true for all the test cases. It is better to set the default to false and create a separate test case for this value to be true. In this way it is more clear to see what changes are introduced and whether the old codegen is affected.

nickarocho:
> Overall, I have concerns about the backward compatibility.
> Suggestion is that instead of changing the behavior of the existing function, maybe create a new function that calls the old one with additional new behaviors.
> For example, instead of changing "generate" function, leave the old generate function as is and create a new one "generateWithTypes" that calls the original generate and then create more objects to be returned along with the results from the original.

Hi @frimfram this is a good call. I'm just wondering if we take this approach, would we need to wrap the `visitor.generate();` in `plugin.ts` (and other `.generate()` calls like the test files) with a conditional check for the feature flag to call the appropriate method?

i.e.
```js
if (this.config.isTimestampFieldsAdded) {
  visitor.generateWithTypes();
} else {
  visitor.generate();
}
```

Thanks for clarifying.

AaronZyLee:
> > Overall, I have concerns about the backward compatibility.
> > Suggestion is that instead of changing the behavior of the existing function, maybe create a new function that calls the old one with additional new behaviors.
> > For example, instead of changing "generate" function, leave the old generate function as is and create a new one "generateWithTypes" that calls the original generate and then create more objects to be returned along with the results from the original.
> 
> Hi @frimfram this is a good call. I'm just wondering if we take this approach, would we need to wrap the `visitor.generate();` in `plugin.ts` (and other `.generate()` calls like the test files) with a conditional check for the feature flag to call the appropriate method?
> 
> i.e.
> 
> ```js
> if (this.config.isTimestampFieldsAdded) {
>   visitor.generateWithTypes();
> } else {
>   visitor.generate();
> }
> ```
> 
> Thanks for clarifying.

@nickarocho I would suggest to limit the scope to the `appsync-javascript/typescript-visitor.ts` rather than `plugin.ts` since the latter one is used by more than javascript generators. An example is like put `generateWithTypes` within `generate`

AaronZyLee (approved):
LGTM 🎲 !

PR #209, opened 2021-07-28
fix(graphql-types-generator): add wrapperr for sub response in angular

### LINKED ISSUE #207: Bug in Angular codegen for GraphQL API subscriptions
refer 
https://github.com/aws-amplify/amplify-codegen/issues/25
https://github.com/aws-amplify/amplify-cli/issues/5284

creating new as all the tickets are closed and the issue still persist.

AaronZyLee:
#### Description of changes

Add `Subscription` type in angular type-gen and the corresponding `Pick` wrapper in subscription response

#### Issue #, if available

fix #207

`yarn test`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/graphql-types-generator/src/angular/index.ts
+++ b/packages/graphql-types-generator/src/angular/index.ts
@@ -14,6 +14,7 @@ import {
 import { typeNameFromGraphQLType } from '../typescript/types';
 import { Property, interfaceDeclaration } from '../typescript/language';
 import { isList } from '../utilities/graphql';
+import { propertyDeclarations } from '../flow/codeGeneration';
 
 export function generateSource(context: LegacyCompilerContext) {
   const generator = new CodeGenerator<LegacyCompilerContext>(context);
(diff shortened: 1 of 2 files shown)


packages/graphql-types-generator/src/angular/index.ts
```diff
@@ -54,6 +52,38 @@ function generateTypes(generator: CodeGenerator, context: LegacyCompilerContext)
...
 
+function generateSubscriptionResponseWrapper(generator: CodeGenerator) {
+  generator.printOnNewline(`
+  export interface SubscriptionResponse<T> {
+    value: GraphQLResult<T>;
+  }`);
+  generator.printNewline();
+}
+function generateSubscriptionOperationTypes(generator: CodeGenerator, context: LegacyCompilerContext) {
+  const typeName = 'Subscription';
```
phani-srikar, line 63: Introducing `Subscription` type might have high probability to clash with developer's existing type with same name. Can we use` __Subscription` or `__SubscriptionContainer` instead?

phani-srikar (approved):
Suggested naming change to avoid conflicts. This should also probably go behind a FF? We can keep the FF deprecation shorter than others, probably a couple months.

packages/graphql-types-generator/src/angular/index.ts
AaronZyLee: I think ff is not necessary because it is a fix and the original code does not work.

AaronZyLee: I change the type name to `__SubscriptionContainer` to avoid conflicts

PR #210, opened 2021-07-28
fix(appsync-modelgen-plugin): dart model internal constructor with no internal fields

HuiSF:
#### Issue #, if available

amplify-flutter [#759](https://github.com/aws-amplify/amplify-flutter/issues/759)

1.  Ran `yarn setup-dev`
2. Tested to generate model in Dart with below schema
3. Generated model should have internal constructor as below example

```graphql
type IDOnlyModel @model {
  id: ID!
}
```

Generated model internal constructor
```dart
// Before the fix
const IDOnlyModel._internal({required this.id}): ;

// After the fix
const IDOnlyModel._internal({required this.id});
```
#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-dart-visitor.test.ts.snap
+++ b/packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-dart-visitor.test.ts.snap
@@ -3355,6 +3355,105 @@ class _authorBookModelType extends ModelType<authorBook> {
 "
 `;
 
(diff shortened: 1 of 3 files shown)


AaronZyLee (changes requested):
The change looks good. Could you add one unit test for this case?

HuiSF:
> The change looks good. Could you add one unit test for this case?

Sure. Update the PR with new unit test to cover the use case of a model has only `id` field.

AaronZyLee (approved):

PR #218, opened 2021-08-12
fix(appsync-modelgen-plugin): add type cast for float fromJson

### LINKED ISSUE #202: Flutter Codegen not handling floats correctly
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

curl

### If applicable, what version of Node.js are you using?

_No response_

### Amplify CLI Version

5.1.2

### What operating system are you using?

Windows (amplify cli on WSL)

### Amplify Categories

auth, storage, function, api

### Amplify Commands

codegen, pull

### Describe the bug

The automatically generated flutter model files do not handle correctly floats and nullable floats in the `.fromJson` function.
Linked issue in [aws-amplify/amplify-flutter](https://github.com/aws-amplify/amplify-flutter/issues/400) 

The current generated models treat a float as:

```dart
   totalValue = json['totalValue'],
```

### Expected behavior

The generated model should treat floats / nullable floats in the `.fromJson` function as:
```dart
totalValue = json['totalValue'] != null
            ? (json['totalValue'] as num).toDouble()
            : json['totalValue'],
```

### Reproduction steps

1. Create a new model with nullable floats within admin UI.
2. Create new registers (one leaving the float null and another one with an integer number)
3. Pull the changes and generate the models.
4. Try to query the models.

### GraphQL schema(s)

<details>

```graphql
# Put schemas below this line


```

</details>

### Log output

<details>

```
# Put your logs below this line


```

</details>

### Additional information

_No response_

AaronZyLee:
#### Description of changes

Add the following change to the float type in `fromJson` method of flutter modelgen generated code
```Dart
floatVal = (json['floatlVal'] as num?)?.toDouble();
```

#### Issue #, if available

fix #202 

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/appsync-modelgen-plugin/src/visitors/appsync-dart-visitor.ts
+++ b/packages/appsync-modelgen-plugin/src/visitors/appsync-dart-visitor.ts
@@ -300,11 +300,9 @@ export class AppSyncModelDartVisitor<
       : `{${model.fields.map(f => `${this.isFieldRequired(f) ? '@required ' : ''}this.${this.getFieldName(f)}`).join(', ')}}`;
     const internalFields = model.fields.filter(f => this.getFieldName(f) !== 'id');
     const internalImpl = this.isNullSafety()
-      ? internalFields.length ? `: ${
-        internalFields
-          .map(f => `_${this.getFieldName(f)} = ${this.getFieldName(f)}`)
-          .join(', ')
-      };` : ';'
+      ? internalFields.length
+        ? `: ${internalFields.map(f => `_${this.getFieldName(f)} = ${this.getFieldName(f)}`).join(', ')};`
+        : ';'
       : ';';
     declarationBlock.addClassMethod(`${this.getModelName(model)}._internal`, '', [{ name: args }], internalImpl, {
       const: true,
@@ -468,10 +466,12 @@ export class AppSyncModelDartVisitor<
                 ),
                 indent(`.toList()`, 2),
                 indent(`: null`),
-              ].filter((e) => e !== undefined).join('\n');
+              ]
+                .filter(e => e !== undefined)
+                .join('\n');
             }
             return [
-              `${fieldName} = json['${varName}']${this.isNullSafety() ? `?['serializedData']`:''} != null`,
+              `${fieldName} = json['${varName}']${this.isNullSafety() ? `?['serializedData']` : ''} != null`,
               indent(
                 `? ${this.getNativeType(field)}.fromJson(new Map<String, dynamic>.from(json['${varName}']${
                   this.isNullSafety() ? `['serializedData']` : ''
@@ -510,6 +510,10 @@ export class AppSyncModelDartVisitor<
               return field.isList
                 ? `${fieldName} = (json['${varName}'] as List<dynamic>)?.map((dynamic e) => e is double ? e.toInt() : e as int)?.toList()`
                 : `${fieldName} = json['${varName}']`;
+            case this.scalars['Float']:
+              return field.isList
+                ? `${fieldName} = json['${varName}']?.cast<${this.getNativeType({ ...field, isList: false })}>()`
+                : `${fieldName} = (json['${varName}'] as num?)?.toDouble()`;
             default:
               return field.isList
                 ? `${fieldName} = json['${varName}']?.cast<${this.getNativeType({ ...field, isList: false })}>()`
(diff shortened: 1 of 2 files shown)


packages/appsync-modelgen-plugin/src/visitors/appsync-dart-visitor.ts
```diff
@@ -510,6 +510,10 @@ export class AppSyncModelDartVisitor<
               return field.isList
                 ? `${fieldName} = (json['${varName}'] as List<dynamic>)?.map((dynamic e) => e is double ? e.toInt() : e as int)?.toList()`
                 : `${fieldName} = json['${varName}']`;
+            case this.scalars['Float']:
+              return field.isList
+                ? `${fieldName} = json['${varName}']?.cast<${this.getNativeType({ ...field, isList: false })}>()`
+                : `${fieldName} = (json['${varName}'] as num?)?.toDouble()`;
```
phani-srikar, line 516: The null check pattern looks inconsistent across the cases. For example, in case of Timestamps, it is `json['${varName}'] != null ? ${fieldNativeType}.fromSeconds(json['${varName}']) : null`

phani-srikar: why is this explicit casting to `num` and then to `Double` necessary?

phani-srikar (changes requested):
please confirm the changes with Flutter team and also verify if this is a breaking change that needs to be backed by Feature Flag.

packages/appsync-modelgen-plugin/src/visitors/appsync-dart-visitor.ts
AaronZyLee: This line was taken from the comment of the original issue. Need to confirm with Flutter team

cshfang: Hmm, rather than casting the list type (which is the default case last resort), I think we should be able to keep the same pattern as most of these other scalars. Since I noticed customers also having trouble with int types, I suggest improving the `Int` case and basically duplicating it for `Float`:
```dart
case this.scalars['Int']:
  return field.isList
    ? `${fieldName} = (json['${varName}'] as List?)?.map((e) => (e as num).toInt()).toList()`
    : `${fieldName} = (json['${varName}'] as num?)?.toInt()`;
case this.scalars['Float']:
  return field.isList
    ? `${fieldName} = (json['${varName}'] as List?)?.map((e) => (e as num).toDouble()).toList()`
    : `${fieldName} = (json['${varName}'] as num?)?.toDouble()`;
```
Technically, we may not need to cast to `num` first since it the compiler would just treat it as dynamic but the cast helps it understand that `toInt()` and `toDouble()` are indeed member functions

cshfang (changes requested):

fjnoyp (approved):

Lorenzohidalgo:
This issue is still happening in Amplify CLI version 5.4.0, at least for non null-safe projects.

Lorenzohidalgo:
just updated to 5.5.0 and it's still happening...

HuiSF:
@Lorenzohidalgo Could you please elaborate "still happening", is that the cli cannot generate model classes containing the fix, or something else?

If the former, could please double check the amplify-codegen version installed in your environment? You can get check the version by running:
```
npm -g list --depth=1
```
Look for `amplify-codegen` it should be `2.26.4`.

Lorenzohidalgo:
> @Lorenzohidalgo Could you please elaborate "still happening", is that the cli cannot generate model classes containing the fix, or something else?
> 
> If the former, could please double check the amplify-codegen version installed in your environment? You can get check the version by running:
> 
> ```
> npm -g list --depth=1
> ```
> 
> Look for `amplify-codegen` it should be `2.26.4`.

Exactly, I Updated the cli to v5.5.0 and the generated files do not contain the fix. The npm command returns an error: `/mnt/c/Program Files/nodejs/npm: 6: Syntax error: word unexpected (expecting "in")`

I'm running amplify cli on WSL.

HuiSF:
Sorry I'm not familiar with running npm in Windows OS.
Could you please try to remove amplify Cli completely and reinstall it?
Please also take a look at [this comment](https://github.com/aws-amplify/amplify-flutter/issues/801#issuecomment-909706509) it was a similar issue.

PR #227, opened 2021-08-18
fix(appsync-modelgen-plugin): remove unnecessary null-aware operators

cshfang:
#### Description of changes
This PR removes unnecessary null-aware operators from generated Dart models
* remove null-aware operator from toList() where already short-circuited
* update time and int scalar type lists to be cast as nullable
* specify has many element as nullable when mapping to JSON

```dart
// current generated model
Map<String, dynamic> toJson() => {
  'id': id, 'todos': _todos?.map((e) => e?.toJson())?.toList(), 'name': _name
};      //                               ^          ^
        // assumes element is not nullable          unnecessary due to short-circuiting
```
```dart
Map<String, dynamic> toJson() => {
  'id': id, 'todos': _todos?.map((Todo? e) => e?.toJson()).toList(), 'name': _name
};
```

#### Issue #, if available
https://github.com/aws-amplify/amplify-flutter/issues/801

* Verified that snapshot errors caused by changes were intentional
* Verified that updated snapshots are correct
* Made sure `yarn build` and `yarn test` ran successfully
* Tested relevant schemas in a Flutter app locally

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-dart-visitor.test.ts.snap
+++ b/packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-dart-visitor.test.ts.snap
@@ -356,21 +356,21 @@ class TemporalTimeModel extends Model {
         timestamp = json['timestamp'] != null
             ? TemporalTimestamp.fromSeconds(json['timestamp'])
             : null,
-        dateList = (json['dateList'] as List)
+        dateList = (json['dateList'] as List?)
             ?.map((e) => TemporalDate.fromString(e))
-            ?.toList(),
-        timeList = (json['timeList'] as List)
+            .toList(),
+        timeList = (json['timeList'] as List?)
             ?.map((e) => TemporalTime.fromString(e))
-            ?.toList(),
-        dateTimeList = (json['dateTimeList'] as List)
+            .toList(),
+        dateTimeList = (json['dateTimeList'] as List?)
             ?.map((e) => TemporalDateTime.fromString(e))
-            ?.toList(),
-        timestampList = (json['timestampList'] as List)
+            .toList(),
+        timestampList = (json['timestampList'] as List?)
             ?.map((e) => TemporalTimestamp.fromSeconds(e))
-            ?.toList(),
-        intList = (json['intList'] as List<dynamic>)
-            ?.map((dynamic e) => e is double ? e.toInt() : e as int)
-            ?.toList();
+            .toList(),
+        intList = (json['intList'] as List?)
+            ?.map((e) => e is double ? e.toInt() : e as int)
+            .toList();
 
   Map<String, dynamic> toJson() => {
         'id': id,
@@ -646,13 +646,13 @@ class TestEnumModel extends Model {
         'id': id,
         'enumVal': enumToString(enumVal),
         'nullableEnumVal': enumToString(nullableEnumVal),
-        'enumList': enumList?.map((e) => enumToString(e))?.toList(),
+        'enumList': enumList?.map((e) => enumToString(e)).toList(),
         'enumNullableList':
-            enumNullableList?.map((e) => enumToString(e))?.toList(),
+            enumNullableList?.map((e) => enumToString(e)).toList(),
         'nullableEnumList':
-            nullableEnumList?.map((e) => enumToString(e))?.toList(),
+            nullableEnumList?.map((e) => enumToString(e)).toList(),
         'nullableEnumNullableList':
-            nullableEnumNullableList?.map((e) => enumToString(e))?.toList()
+            nullableEnumNullableList?.map((e) => enumToString(e)).toList()
       };
 
   static final QueryField ID = QueryField(fieldName: \\"testEnumModel.id\\");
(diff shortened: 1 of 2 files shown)


packages/appsync-modelgen-plugin/src/visitors/appsync-dart-visitor.ts
```diff
@@ -300,11 +300,9 @@ export class AppSyncModelDartVisitor<
...
     const internalFields = model.fields.filter(f => this.getFieldName(f) !== 'id');
     const internalImpl = this.isNullSafety()
-      ? internalFields.length ? `: ${
-        internalFields
-          .map(f => `_${this.getFieldName(f)} = ${this.getFieldName(f)}`)
-          .join(', ')
-      };` : ';'
+      ? internalFields.length
+        ? `: ${internalFields.map(f => `_${this.getFieldName(f)} = ${this.getFieldName(f)}`).join(', ')};`
+        : ';'
```
cshfang, line 305: Auto-updated by lint

packages/appsync-modelgen-plugin/src/visitors/appsync-dart-visitor.ts
```diff
@@ -468,10 +466,12 @@ export class AppSyncModelDartVisitor<
                 ),
                 indent(`.toList()`, 2),
                 indent(`: null`),
-              ].filter((e) => e !== undefined).join('\n');
+              ]
+                .filter(e => e !== undefined)
+                .join('\n');
```
cshfang, line 471: Auto-updated by lint

packages/appsync-modelgen-plugin/src/visitors/appsync-dart-visitor.ts
```diff
@@ -468,10 +466,12 @@ export class AppSyncModelDartVisitor<
...
             }
             return [
-              `${fieldName} = json['${varName}']${this.isNullSafety() ? `?['serializedData']`:''} != null`,
+              `${fieldName} = json['${varName}']${this.isNullSafety() ? `?['serializedData']` : ''} != null`,
```
cshfang, line 474: Auto-updated by lint

packages/appsync-modelgen-plugin/src/visitors/appsync-dart-visitor.ts
```diff
@@ -532,13 +532,14 @@ export class AppSyncModelDartVisitor<
         const fieldName = `${this.isNullSafety() && field.name !== 'id' ? '_' : ''}${this.getFieldName(field)}`;
         if (this.isModelType(field)) {
           if (field.isList) {
-            return `'${varName}': ${fieldName}?.map((e) => e?.toJson())?.toList()`;
+            const modelName = this.getNativeType({ ...field, isList: false });
+            return `'${varName}': ${fieldName}?.map((${modelName}? e) => e?.toJson()).toList()`;
```
Jordan-Nelson, line 536: Is the `?` in `map((${modelName}? e)` correct? Can the list have null values?

packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-dart-visitor.test.ts.snap
```diff
@@ -2853,7 +2853,7 @@ class Blog extends Model {
   Map<String, dynamic> toJson() => {
         'id': id,
         'name': name,
-        'posts': posts?.map((e) => e?.toJson())?.toList(),
+        'posts': posts?.map((Post? e) => e?.toJson()).toList(),
```
Jordan-Nelson, line 2856: Can `Post` be null?

packages/appsync-modelgen-plugin/src/visitors/appsync-dart-visitor.ts
cshfang: My understanding is that it is possible with non-required schemas? Either way, I think this is a lower impact change since we're already using the nullable operator with `toJson`

AaronZyLee (commented):
I check out the original issue and the code change is enough to solve the problem. However, I have some thoughts on the change to the list model type.
Considering the following fields:

- `post: [Post]`
- `post: [Post!]`
- `post: [Post]!`
- `post: [Post!]!`

Right now the codegen will convert these fields to all nullable fields in field declaration but with a bug that now converting the element to nullable one. All the fields above will be converted to `final List<Post>? _post`.

If the concept above is correct, one way to fix this issue might be to add nullable to element as well, where the output should be `final List<Post?>? _post`. In this case, we can keep the question mark unchanged for `e` in `toJson` method

packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-dart-visitor.test.ts.snap
AaronZyLee: Post is not nullable when in list, which I think is a bug to fix. More details in my comment below

AaronZyLee (approved):
Discussed offline. The code is sufficient to fix the warning. Tested locally and the change LGTM 🎲

HuiSF (approved):

PR #259, opened 2021-10-13
chore: removing deprecated feature flags, updating references to only reference default implementations, and deprecate reference to cli repo implementations

alharris-at:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


FeatureFlags.md
```diff
@@ -1,27 +1,4 @@
...
-
-3. **useTypesGeneratorPlugin**: When set to `true`, the `graphql-types-generator` implementation in this repository will be used to generate the language specific type definitions from GraphQL schema. When set to `false`, the `amplify-graphql-docs-generator` implementation from [CLI repo](https://github.com/aws-amplify/amplify-cli/tree/master/packages/amplify-graphql-types-generator) will be used to generate the type defintions. 
-
-The package implementations in this repository will be actively maintained going forward while the implementations in [Amplify CLI repository](https://github.com/aws-amplify/amplify-cli/tree/master/packages) will be deprecated on May 1st, 2021. 
-The package implementations in this repository will support all the APIs that are supported by their counterparts in the CLI repository. 
-
-These Feature Flags are set to `true` by default for new projects created using Amplify CLI and `false` for existing projects.
-We recommend setting the above Feature Flags to `true` in your `cli.json` file to consume the package implementations from this repository and take advtange of bug fixes, enhancements etc being made to these generators.
```
phani-srikar, line 20: This might be a useful info to retain as it conveys to the developer that they should work on setting these FFs to `true`. We might additionally also want to callout that each FF has a deprecation date set to give them time to migrate.

packages/graphql-docs-generator/__tests__/generator/generateAllOperations.test.ts
```diff
@@ -69,111 +69,3 @@ describe('generateAllOperations', () => {
...
-        ...mockOperationResult,
-      },
-    ]);
-    expect(generateOperation).toHaveBeenCalledTimes(1);
-    expect(getFields).toHaveBeenCalledTimes(1);
-    expect(generateOperation).toHaveBeenCalledWith('value', mockSchema, maxDepth, generateOptions);
-  });
-});
-
-describe('test capitalizeFirstLetter', () =>{
```
phani-srikar, line 121: I believe this method is still being used. So we might want to retain this test suite.

packages/graphql-docs-generator/__tests__/generator/generateAllOperations.test.ts
```diff
@@ -69,111 +69,3 @@ describe('generateAllOperations', () => {
...
-  it('test capitalize first letter with camelCase string', () => {
-    expect(capitalizeFirstLetter('camelCase')).toEqual('CamelCase');
-  });
-
-  it('test capitalize first letter with PascalCase string', () => {
-    expect(capitalizeFirstLetter('PascalCase')).toEqual('PascalCase');
-  });
-});
-
-describe('test lowerCaseFirstLetter', () =>{
```
phani-srikar, line 151: Retain this suite.

phani-srikar (approved):
Minor changes suggested. Looks good. Some points to consider:
1. We need to remove these deprecated FFs from CLI https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-cli-core/src/feature-flags/featureFlags.ts along with the old packages that are still in their repo. For example, [old modelgen](https://github.com/aws-amplify/amplify-cli/tree/master/packages/amplify-codegen-appsync-model-plugin). We can create a PR and discuss with CLI team if they are okay with the package deletion.
2. I would be more comfortable releasing this as a breaking change on codegen side and CLI side because there could be many existing projects with these FFs set to `false` by default. 
3. Another smoother approach we could take is to not release these code changes yet but first set these FFs to true for all projects in [CLI](https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-cli-core/src/feature-flags/featureFlags.ts), see the customer response and then go ahead with this release. This would be easier to roll back just in case.

packages/graphql-docs-generator/__tests__/generator/generateAllOperations.test.ts
alharris-at: Same, will add this back in.

FeatureFlags.md
alharris-at: > Another smoother approach we could take is to not release these code changes yet but first set these FFs to true for all projects in CLI, see the customer response and then go ahead with this release. This would be easier to roll back just in case.

That probably makes some sense. We can drop it and revisit it in a month or two.

I think I'll take that approach, and just go with the CLI/docs update.

PR #268, opened 2021-10-19
generate searchable aggregateItems 2 additional levels

sundersc:
#### Description of changes
GraphQL searchable aggregateItems field contains a union type for the result field. It requires two additional level of traversing with the default statement depth (2) to generate the queries as expected.

- Manual test
- yarn test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/graphql-docs-generator/src/generator/getFields.ts
+++ b/packages/graphql-docs-generator/src/generator/getFields.ts
@@ -67,6 +67,8 @@ function adjustDepth(field, depth) {
   const maxDepth = 100;
   if (isGraphQLAggregateField(field) && depth < maxDepth) {
     return depth + 1;
+  } else if (depth >= maxDepth) {
+    throw new Error('Statement generation depth exceeded the maximum allowed limit');
   }
   return depth - 1;
 }


packages/graphql-docs-generator/src/generator/getFields.ts
```diff
@@ -61,3 +61,10 @@ export default function getFields(
     hasBody: !!(fields.length || fragments.length),
   };
 }
+
+function adjustDepth(fieldName, depth) {
+  if (fieldName == 'aggregateItems') {
+    return depth + 1;
```
frimfram, line 67: can the depth be increased indefinitely?  Would need to check for some upper bound.

sundersc: I've added a statement to check for the typename. Since 'SearchableAggregateResult' type is generated by CLI V2 transformer, we can be sure this doesn't go through infinite loop.

frimfram: in the case of isGraphQLAggregateField(field) is true and depth >= maxDepth, we may need to either return depth or throw errorr.  Currently it'll return depth-1 which may not be what's intended.

AaronZyLee: Same question as above. If it is the aggregate field but the depth reaches maxDepth, we cannot return `depth-1` here. I suggest throwing an error.

packages/graphql-docs-generator/__tests__/generator/getFields.test.ts
```diff
@@ -185,4 +185,73 @@ describe('getField', () => {
...
+        },
+      }),
+    });
+
+    it('aggregateItems property should traverse two additional levels to generate required fields with default depth 2', () => {
+      const maxDepth = 2;
+      const getPossibleTypeSpy = jest.spyOn(schema, 'getPossibleTypes');
+      getFields(schema.getQueryType().getFields().aggregateItems, schema, maxDepth, { useExternalFragmentForS3Object: false });
+      expect(getPossibleTypeSpy).toHaveBeenCalled();
+      expect(getFragment).toHaveBeenCalled();
```
AaronZyLee, line 242: The changes happen on `getField` but is not covered in the test

packages/graphql-docs-generator/__tests__/generator/getFields.test.ts
```diff
@@ -185,4 +185,73 @@ describe('getField', () => {
...
+      const commonField = []; // unions don't have to have common field
+
+      expect(getFragment.mock.calls[0][0]).toEqual(aggregateScalarResult);
+      expect(getFragment.mock.calls[0][1]).toEqual(schema);
+      expect(getFragment.mock.calls[0][2]).toEqual(maxDepth - 1);
+      expect(getFragment.mock.calls[0][3]).toEqual(commonField);
+
+      expect(getFragment.mock.calls[1][0]).toEqual(aggregateBucketResult);
+      expect(getFragment.mock.calls[1][1]).toEqual(schema);
+      expect(getFragment.mock.calls[1][2]).toEqual(maxDepth - 1);
```
AaronZyLee, line 253: If the `getField` is tested, the expect value should be `depth+1`

PR #184, opened 2021-06-03
feat(modelgen): ios - emit provider in @auth rules

diegocstn:
#### Description of changes
Currently, models generated for iOS don't have the provider attribute of the @auth rule directive.
This PR adds this attribute as an enum value with the following cases: `.apiKey`, `.oidc`, `.userPools`, `.iam`.

CLI feature flag PR: https://github.com/aws-amplify/amplify-cli/pull/7447
Related Android PR: https://github.com/aws-amplify/amplify-codegen/pull/168

**iOS library PR: https://github.com/aws-amplify/amplify-ios/pull/1258**

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed or added](https://github.com/aws-amplify/amplify-codegen/blob/master/CONTRIBUTING.md#tests)
- [ ] Relevant documentation is changed or added (and PR referenced)
- [x] Breaking changes to existing customers are released behind a feature flag or major version update
- [ ] Changes are tested using sample applications for all relevant platforms (iOS/android/flutter/Javascript) that use the feature added/modified

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #272, opened 2021-10-26
feat(appsync-dart-visitor): insert auth provider info

HuiSF:
# Please do not merge for now

This change inserts `provider` information to `AuthRule` initializer in generated Dart model classes. This changes is required to align with the multi-auth implementation in amplify-ios (https://github.com/aws-amplify/amplify-codegen/pull/184).

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-dart-visitor.test.ts
+++ b/packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-dart-visitor.test.ts
@@ -15,12 +15,13 @@ const getVisitor = (
   enableDartNullSafety: boolean = false,
   enableDartNonModelGeneration: boolean = true,
   isTimestampFieldsAdded: boolean = false,
+  emitAuthProvider: boolean = false
 ) => {
   const ast = parse(schema);
   const builtSchema = buildSchemaWithDirectives(schema);
   const visitor = new AppSyncModelDartVisitor(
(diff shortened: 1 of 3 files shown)


packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-dart-visitor.test.ts
```diff
@@ -15,12 +15,13 @@ const getVisitor = (
   enableDartNullSafety: boolean = false,
   enableDartNonModelGeneration: boolean = true,
   isTimestampFieldsAdded: boolean = false,
+  emitAuthProvider: boolean = false
```
dnys1, line 18: What would be the case we don’t want to emit them?

HuiSF: For all existing projects the feature flag will be set as false, for all new projects this feature flag value is set as true per amplify-ios library maintainers' input.

dnys1: Hmm I think iOS's situation was a bit different. I would prefer to have it always on, but I will default to your judgment since you had the conversations with them.

dnys1 (approved):

PR #274, opened 2021-10-30
fix(graphql): aggregate-code-fix

sundersc:
#### Description of changes
Fix for GraphQL aggregateItems datatype change.

- Manual test
- Unit test

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/graphql-docs-generator/src/generator/getFields.ts
+++ b/packages/graphql-docs-generator/src/generator/getFields.ts
@@ -77,9 +77,16 @@ function isGraphQLAggregateField(field) {
   if (
     field &&
     field.name == 'aggregateItems' &&
-    field.type?.ofType?.name == 'SearchableAggregateResult'
+    getBaseType(field.type) == 'SearchableAggregateResult'
   ) {
     return true;
   }
   return false;
+}
+
+function getBaseType(field) {
+  if(field && field.ofType) {
+    return getBaseType(field.ofType);
+  }
+  return field?.name;
 }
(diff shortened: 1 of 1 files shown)


packages/graphql-docs-generator/src/generator/getFields.ts
```diff
@@ -77,9 +77,16 @@ function isGraphQLAggregateField(field) {
...
     field.name == 'aggregateItems' &&
-    field.type?.ofType?.name == 'SearchableAggregateResult'
+    getBaseType(field.type) == 'SearchableAggregateResult'
   ) {
     return true;
   }
   return false;
+}
+
+function getBaseType(field) {
```
phani-srikar, line 87: can we please rename the argument to `type` instead of `field` as we are actually passing on the type of the field.

phani-srikar (changes requested):
Correct me if I am wrong: The problem we are trying to solve here is that there can be multiple `ofType` properties on `field.type` and `SearchableAggregateResult` should be compared to the base type not the top level `type.ofType`. If this is the case the changes look good to me. 
Can we have atleast one simple unit test to cover this case? Thanks

packages/graphql-docs-generator/src/generator/getFields.ts
sundersc: Updated. The existing unit test case takes care of this scenario. 
https://github.com/aws-amplify/amplify-codegen/blob/master/packages/graphql-docs-generator/__tests__/generator/getFields.test.ts#L189-L257

The change here is to treat `[SearchableAggregateResult]`, `[SearchableAggregateResult]!` and `[SearchableAggregateResult!]!` in the same way.

phani-srikar (approved):
discussed offline. LGTM!

PR #276, opened 2021-11-01
fix(amplify-codegen): fix v2 has one process connection directive

lazpavel:
#### Description of changes
- remove field from model for hasOne directive

- manually tested an datastore app with hasOne and v2 transformer

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
+++ b/packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
@@ -108,7 +108,7 @@ export interface RawAppSyncModelConfig extends RawConfig {
    * @type boolean
    * @descriptions optional boolean which generates the list types to respect the nullability as defined in the schema
    */
-   handleListNullabilityTransparently?: boolean;
+  handleListNullabilityTransparently?: boolean;
   /**
    * @name usePipelinedTransformer
    * @type boolean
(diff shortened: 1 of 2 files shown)


packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
```diff
@@ -652,7 +669,7 @@ export class AppSyncModelVisitor<
       model.fields.forEach(field => {
         const connectionInfo = processConnectionsV2(field, model, this.modelMap);
         if (connectionInfo) {
-          if (connectionInfo.kind === CodeGenConnectionType.HAS_MANY || connectionInfo.kind === CodeGenConnectionType.HAS_ONE) {
+          if (connectionInfo.kind === CodeGenConnectionType.HAS_MANY) {
```
phani-srikar, line 672: The `HAS_ONE` connection no longer requires the connected field? Can you explain why? (possibly also add a unit test)

phani-srikar (changes requested):

packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
lazpavel: removed the connected field from transformer V2 to match the behavior for the same schema in V1 (codegen seems to consider @connection for the equivalent V1 schema as a belongsTo connection type)
added a unit test for the change

phani-srikar (approved):
Thanks. LGTM!

PR #277, opened 2021-11-03
feat(amplify-codegen): change flutter import from datastore to core

ragingsquirrel3:
This PR changes amplify-flutter (dart) codegen models so that they import from the core plugin instead of the datastore_plugin_interface. This is part of an effort in the next major version of amplify-flutter to allow customers to use codegen models without taking on a datastore dependency. One of the features in this change is model-based GraphQL helpers for the API category. That feature will allow users to have codegen files in an app that uses the API category without the datastore category. For that reason, codegen dependencies should come from core so they can be used with datastore or API exclusively. There is an amplify-flutter change to make those available in core https://github.com/aws-amplify/amplify-flutter/pull/1023.

This codegen change is desired to be made available via a preview release so customers can use it during the preview stage of amplify-flutter 0.3.0 release (0.3.0-rc.x). For that reason, this PR is against the tagged-release/custom-type branch so it can use the same preview tag (presumably with incremented number). These changes are intended for use with same major version (and preview releases) of amplify-flutter.

In codegen repo (on branch for this PR): ran `setup-dev` and `yarn setup-dev` to make local changes available.

In amplify-flutter (on branch with deps moved), went to datastore example app and ran `amplify-dev codegen models`. With locally updated codegen models, ran the integration tests on android and ios, as well as flutter unit tests.

In amplify-flutter, in separate branch for model helpers, ran the same codegen command in API example app without datastore (to test API only use). flutter unit tests and integration tests pass there.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x ] Breaking changes to existing customers are released behind a feature flag or major version update

### DIFF
--- a/packages/amplify-codegen/src/commands/models.js
+++ b/packages/amplify-codegen/src/commands/models.js
@@ -86,7 +86,7 @@ async function generateModels(context) {
 
   const generateIndexRules = readFeatureFlag('codegen.generateIndexRules');
   const emitAuthProvider = readFeatureFlag('codegen.emitAuthProvider');
-  const usePipelinedTransformer = readFeatureFlag('graphQLTransformer.useExperimentalPipelinedTransformer');
+  const usePipelinedTransformer = readFeatureFlag('graphQLTransformer.useExperimentalPipelinedTransformer')
   const transformerVersion = readNumericFeatureFlag('graphQLTransformer.transformerVersion');
 
   let enableDartNullSafety = readFeatureFlag('codegen.enableDartNullSafety');
--- a/packages/amplify-codegen/src/commands/models.js
+++ b/packages/amplify-codegen/src/commands/models.js
@@ -30,6 +30,19 @@ const readFeatureFlag = key => {
   return flagValue;
 };
 
+/**
+ * Returns feature flag value, default to `1`
+ * @param {string} key feature flag id
+ * @returns
+ */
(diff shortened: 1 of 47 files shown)


packages/amplify-codegen/src/commands/models.js
```diff
@@ -86,7 +86,7 @@ async function generateModels(context) {
 
   const generateIndexRules = readFeatureFlag('codegen.generateIndexRules');
   const emitAuthProvider = readFeatureFlag('codegen.emitAuthProvider');
-  const usePipelinedTransformer = readFeatureFlag('graphQLTransformer.useExperimentalPipelinedTransformer')
+  const usePipelinedTransformer = readFeatureFlag('graphQLTransformer.useExperimentalPipelinedTransformer');
```
ragingsquirrel3, line 89: seems like some of these files got reformatted unrelated to my change, lmk if prefer to not have in this PR

AaronZyLee: It is better not to include the lint change since it is too many or you can comment on the non-lint change so that I can locate it easily

ragingsquirrel3: ok will remove, thanks

AaronZyLee (approved):
LGTM 🍵

PR #304, opened 2021-11-16
fix: repo dev dependencies of cli components

AaronZyLee:
#### Description of changes

The recent release of CLI has major version bump for its component packages, which codegen repo has a dev dependency. Bump the dependency to next major version so that the `amplify-dev` will not break

Related packages:

- `amplify-cli-core`: `1.x` to `2.x`
- `graphql-transformer-core`: `6.x` to `7.x`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,9 +95,9 @@
     "ts-jest": "^25.5.0",
     "ts-node": "^8.10.1",
     "typescript": "^3.8.3",
-    "amplify-cli-core": "^1.17.2",
-    "graphql-transformer-core": "^6.26.2",
-    "amplify-headless-interface": "^1.6.0"
+    "amplify-cli-core": "^2.3.0",
+    "graphql-transformer-core": "^7.2.1",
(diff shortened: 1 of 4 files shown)


alharris-at (approved):

phani-srikar (changes requested):
We might want to upgrade these CLI package dependencies listed as `peerDependencies` in other places like inside `amplify-codegen, amplify-codegen-e2e-core and amplify-codegen-e2e-tests` too.

alharris-at:
> so that the `amplify-dev` will not break

Did this package not break based on these as well? If not they may not be proper dev deps.

AaronZyLee:
> > so that the `amplify-dev` will not break
> 
> Did this package not break based on these as well? If not they may not be proper dev deps.

Only dev deps but not for production

phani-srikar (approved):
LGTM!

PR #307, opened 2021-11-16
fix(amplify-codegen): support for @default directive for required fields

lazpavel:
#### Description of changes
- support for @default directive required fields

- `yarn test` passes

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-dart-visitor.test.ts
```diff
@@ -346,6 +346,18 @@ describe('AppSync Dart Visitor', () => {
       const generatedCode = visitor.generate();
       expect(generatedCode).toMatchSnapshot();
     });
+
+    it('Should set fields with @default as nullable', () => {
```
marcvberg, line 350: Is this the behavior we want vs erroring out? If the schema shows it as non nullable, it could be misleading

packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
```diff
@@ -788,7 +788,16 @@ export class AppSyncModelVisitor<
...
+    const hasDefaultDirective = this.hasDefaultDirective(field);
+    return !(this.config.handleListNullabilityTransparently
+      ? field.isList
+        ? field.isListNullable
+        : field.isNullable || hasDefaultDirective
+      : field.isNullable || hasDefaultDirective);
+  }
+
+  protected hasDefaultDirective(field: CodeGenField): boolean {
+    return !!field.directives.find(directive => directive.name === 'default');
```
marcvberg, line 800: You can use `field.directives.some` rather than find if you just want to get a boolean result back, then we don't need the `!!`

packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-dart-visitor.test.ts
lazpavel: yes, this is similar to how we handle the ID fields

marcvberg (approved):
LGTM

cjihrig:
ping @aws-amplify/amplify-codegen

AaronZyLee (commented):
The change looks good but is not sufficient to accomplish the goal of using `@default` in datastore modelgen. The PR only addresses converting the nullability the field with default but not assigning the values to it. 
Besides, when I run the current modelgen command in production on schema with `@default`, I don't find any errors blocking the code generation. I am concerned the introduction of the insufficient implementation might have impact on current user experience.
The expected behavior should be like when I use modelgen commands with `@default` schema, I have the model code generated which enable user to run `Datastore.save()` without including the default field even if it is required in my schema and the value is automatically assigned with the directive value.

Another confusion of this practice is that how we treat these required fields with `@default`. Are they in fact a non-required field (no difference whether there is an `!` mark) or they are actually a required field with value assigned when we do not provide value in the creation input

alex-vladut:
Hello @lazpavel @AaronZyLee, I see this PR is stuck for quite some time now. Is anyone able to provide any ETA in case it is still going to be addressed at some point? Thanks

alharris-at:
Need to close as a team whether or not we want to have the platforms fill in default values in datastore. Transferring ownership of this PR from @lazpavel to the the @aws-amplify/amplify-data team.

We can either keep required, and add default value, or remove the required constraint locally, and allow the cloud to sync.

PR #310, opened 2021-11-20
Fix: stop unnecessarily removing fields from belongsTo models

marcvberg:
#### Description of changes

A remnant of the V1 transformer is removing any auto generated fields from a belongsTo type connection. This was done to patch a bug in V1 that was solved by the explicit V2 directives, so we're removing the part where any belongsTo creates a field, as well as the part where added fields from belongsTo are removed.

Running unit tests and running codegen locally

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/appsync-modelgen-plugin/src/utils/process-belongs-to.ts
+++ b/packages/appsync-modelgen-plugin/src/utils/process-belongs-to.ts
@@ -45,13 +45,15 @@ export function processBelongsToConnection(
   //  track the connection and that field is not part of the selection set
   // but if the field are connected using fields argument in connection directive
   // we are reusing the field and it should be preserved in selection set
-  const isConnectingFieldAutoCreated = connectionFields.length === 0;
+  const otherSideHasMany = otherSideField.isList;
+  const isConnectingFieldAutoCreated = false;
 
   return {
     kind: CodeGenConnectionType.BELONGS_TO,
     connectedModel: otherSide,
     isConnectingFieldAutoCreated,
-    targetName: connectionFields[0] || makeConnectionAttributeName(model.name, field.name),
+    targetName: connectionFields[0] || (otherSideHasMany ? makeConnectionAttributeName(otherSide.name, otherSideField.name) :
+      makeConnectionAttributeName(model.name, field.name)),
   };
 }
 
--- a/packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
+++ b/packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
@@ -693,9 +693,6 @@ export class AppSyncModelVisitor<
               isList: false,
               isNullable: field.isNullable,
             });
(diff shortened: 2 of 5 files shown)


packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
```diff
@@ -693,9 +693,6 @@ export class AppSyncModelVisitor<
               isList: false,
               isNullable: field.isNullable,
             });
-          } else if (connectionInfo.targetName !== 'id') {
-            // Need to remove the field that is targetName
-            removeFieldFromModel(model, connectionInfo.targetName);
```
frimfram, line 698: Don't we still need to remove if it's not 'id'?

packages/appsync-modelgen-plugin/src/utils/process-belongs-to.ts
```diff
@@ -45,13 +45,15 @@ export function processBelongsToConnection(
...
-  const isConnectingFieldAutoCreated = connectionFields.length === 0;
+  const otherSideHasMany = otherSideField.isList;
+  const isConnectingFieldAutoCreated = false;
 
   return {
     kind: CodeGenConnectionType.BELONGS_TO,
     connectedModel: otherSide,
     isConnectingFieldAutoCreated,
-    targetName: connectionFields[0] || makeConnectionAttributeName(model.name, field.name),
+    targetName: connectionFields[0] || (otherSideHasMany ? makeConnectionAttributeName(otherSide.name, otherSideField.name) :
```
AaronZyLee, line 55: Just want to confirm that whether the targetName is generated correctly in a `belongsTo` vs `hasOne` situation since it is opposite with the `belongsTo` vs `hasMany`. Otherwise LGTM

packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
marcvberg: The reason I'm not removing anymore here is because belongsTo will never add fields anymore because of my change, and this breaks logic because hasMany will add its connecting field on the connected model. So when hasMany and belongsTo both reference that connected field, this code would snip it out even though it's necessary. This behavior prevents belongsTo from working correctly with DataStore

AaronZyLee: I think V2 transformer uses different approach and there is no need to remove non-id target fields

packages/appsync-modelgen-plugin/src/utils/process-belongs-to.ts
marcvberg: This is correct for hasMany, and I only changed the naming behavior for hasMany since that's what the bug was in response to, nothing was reported for hasOne

AaronZyLee: Although it is sufficient for solving the sev2

PR #319, opened 2021-12-03
fix: partially rolling back commit 310 for flutter, swift, and java platforms

alharris-at:
#### Description of changes
Working on high severity issue, we believe a recent commit https://github.com/aws-amplify/amplify-codegen/pull/310 has caused a regression on swift, dart, and android platforms. This is an attempt to revert that commit only for those platforms.

### Modelgen:
- Reverted removing targetName in Java, swift and dart
- Used field with model type instead of auto generated ID in `associatedWith` in swift and dart

#### Issue #, if available
N/A

 `yarn test`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/appsync-modelgen-plugin/src/utils/process-connections-v2.ts
+++ b/packages/appsync-modelgen-plugin/src/utils/process-connections-v2.ts
@@ -12,6 +12,7 @@ export function getConnectedFieldV2(
   model: CodeGenModel,
   connectedModel: CodeGenModel,
   directiveName: string,
+  shouldUseModelNameFieldInHasManyAndBelongsTo: boolean = false
 ): CodeGenField {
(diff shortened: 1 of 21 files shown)


packages/appsync-modelgen-plugin/src/utils/process-connections-v2.ts
```diff
@@ -12,6 +12,7 @@ export function getConnectedFieldV2(
   model: CodeGenModel,
   connectedModel: CodeGenModel,
   directiveName: string,
+  shouldUseModelNameFieldInHasManyAndBelongsTo: boolean = false
```
alharris-at, line 15: I'd rather this doesn't have a default value so we're explicit.

alharris-at (commented):
One comment, otherwise LGTM.

PR #326, opened 2021-12-09
fix: set associatedWith to correct field with hasOne/hasMany and belongsTo

dpilch:
#### Description of changes

Fix regression when introduced with GQL v2 to set correct `associatedWith` field with `hasOne` and `belongsTo`.

#### Description of how you validated changes

QA Example:

```
type Project9 @model {
  id: ID!
  name: String
  team: Team9 @hasOne
}

type Team9 @model {
  id: ID!
  name: String!
  project: Project9 @belongsTo
}
```

Current incorrect generation:
```
// Incorrectly generated parent model
public final class Project9 implements Model {
  private final @ModelField(targetType="ID", isRequired = true) String id;
  private final @ModelField(targetType="String") String name;
  private final @ModelField(targetType="Team9") @HasOne(associatedWith = "id", type = Team9.class) Team9 team = null;
  private final @ModelField(targetType="ID") String project9TeamId;
}

// Child model is generated correctly, but I'll leave it here for reference
public final class Team9 implements Model {
  private final @ModelField(targetType="ID", isRequired = true) String id;
  private final @ModelField(targetType="String", isRequired = true) String name;
  private final @ModelField(targetType="Project9") @BelongsTo(targetName = "team9ProjectId", type = Project9.class) Project9 project;
}
```

Expected:
```
public final class Project9 implements Model {
  private final @ModelField(targetType="ID", isRequired = true) String id;
  private final @ModelField(targetType="String") String name;
  private final @ModelField(targetType="Team9") @HasOne(associatedWith = "project", type = Team9.class) Team9 team = null;
  private final @ModelField(targetType="ID") String project9TeamId;
}
```

I created a snapshot test to capture this.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed or added](https://github.com/aws-amplify/amplify-codegen/blob/master/CONTRIBUTING.md#tests)
- [ ] Relevant documentation is changed or added (and PR referenced)
- [ ] Breaking changes to existing customers are released behind a feature flag or major version update
- [ ] Changes are tested using sample applications for all relevant platforms (iOS/android/flutter/Javascript) that use the feature added/modified

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #331, opened 2021-12-16
Fix implicit uni-directioanl relationships

phani-srikar:
#### Description of changes

1. Reverts the changes in #326 that caused a regression in JS, Native uni-directional implicit relations as it removes the connected ID field. 
2. AssociatedWith attribute is correctly set to the model field instead of `ID` field for bi-directional implicit relations.
3. Tests are refactored to cover one schema per test for better clarity. 

unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


alharris-at (approved):

alharris-at:
Approving as the change seems sane, but release should definitely still be pending full suite of testing on all platforms.

PR #370, opened 2022-01-13
fix: add amplify prompts in package.json

AaronZyLee:
#### Description of changes

Add `amplify-prompts` as dependency because of old version CLI which does not have this package will break

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-codegen/package.json
+++ b/packages/amplify-codegen/package.json
@@ -37,7 +37,8 @@
     "js-yaml": "^4.0.0",
     "ora": "^4.0.3",
     "semver": "^7.3.5",
-    "slash": "^3.0.0"
+    "slash": "^3.0.0",
+    "amplify-prompts": "^1.6.2"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


dxd1 (approved):

alharris-at (approved):
Can we verify this against the failing case CLI has identified?

packages/amplify-codegen/package.json
```diff
@@ -37,7 +37,8 @@
     "js-yaml": "^4.0.0",
     "ora": "^4.0.3",
     "semver": "^7.3.5",
-    "slash": "^3.0.0"
+    "slash": "^3.0.0",
+    "amplify-prompts": "^1.6.2"
```
phani-srikar, line 41: Shouldn't this be in peer dependencies?

alharris-at: Our code takes a direct dependency on it.

phani-srikar (approved):

PR #380, opened 2022-01-25
chore(amplify-codegen): change flutter import path from datastore to core

ragingsquirrel3:
In flutter(dart) codegen model, change import paths from datastore to core.

Feature flag will be implemented on the target branch flutter-core-dependency-update. As written now, Flutter codegen models with this change will need an upcoming release (0.4.0) of amplify-flutter to resolve correctly. For that reason, not sure if this change should have a feature flag.

#### Description of changes
Changes flutter (dart) codegen so that generated models will import from amplify_core, rather than datastore. This is to support a feature in 0.4.0 amplify-flutter where API users can generate models without a datastore dependency. For this reason, the models need to import from core, as implemented here.

The codegen files here would depend on a version of amplify-flutter with the deps moved (0.4.0+). 0.4.0 will continue to export the same things from datastore to maintain compatibility with older codegen files, but the new codegen files with this change will need a newer version of amplify-flutter to correctly resolve dependencies.

#### Issue #, if available
n/a

In amplify-flutter branch which has necessary dependencies moved to core, ran `amplify-dev codegen models` with the branch from this PR. Ran integration tests and unit tests that import from those files.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Breaking changes to existing customers are released behind a feature flag or major version update (see above)


phani-srikar (approved):
LGTM! As pointed out offline, this feature should get released with a Feature Flag. Relevant changes will be added in the base feature branch.

PR #382, opened 2022-01-28
[Dart] Update imports generated for flutter datastore plugin dependency

phani-srikar:
#### Description of changes

Context as mentioned in [this PR](https://github.com/aws-amplify/amplify-codegen/pull/380):
Changes flutter (dart) codegen so that generated models will import from amplify_core, rather than datastore. This is to support a feature in 0.4.0 amplify-flutter where API users can generate models without a datastore dependency. For this reason, the models need to import from core, as implemented here.

The codegen files here would depend on a version of amplify-flutter with the deps moved (0.4.0+). 0.4.0 will continue to export the same things from datastore to maintain compatibility with older codegen files, but the new codegen files with this change will need a newer version of amplify-flutter to correctly resolve dependencies.

These changes:
- Add a version check to determine the relevant datastore library to import. `v>=0.4.0` Amplify Flutter library users will be able to import the new `amplify_core` package.

Added relevant unit test with snapshots

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-codegen/src/commands/models.js
+++ b/packages/amplify-codegen/src/commands/models.js
@@ -109,7 +109,7 @@ async function generateModels(context) {
     // override isTimestampFieldsAdded to true when using amplify-flutter > 0.3.0 || > 0.3.0-rc.2
     isTimestampFieldsAdded = validateAmplifyFlutterCapableZeroThreeFeatures(projectRoot);
     enableDartZeroThreeFeatures = validateAmplifyFlutterCapableZeroThreeFeatures(projectRoot);
-    // This feature is supported only for users using amplify-flutter > 0.4.0 || > 0.4.0-rc.2
+    // This feature is supported only for users using amplify-flutter > 0.4.0 || > 0.4.0-rc.1
     dartUpdateAmplifyCoreDependency = validateAmplifyFlutterCoreLibraryDependency(projectRoot);
   }
 
--- a/packages/amplify-codegen/src/commands/models.js
+++ b/packages/amplify-codegen/src/commands/models.js
@@ -7,6 +7,7 @@ const gqlCodeGen = require('@graphql-codegen/core');
 const { getModelgenPackage } = require('../utils/getModelgenPackage');
 const { validateDartSDK } = require('../utils/validateDartSDK');
 const { validateAmplifyFlutterCapableZeroThreeFeatures } = require('../utils/validateAmplifyFlutterCapableZeroThreeFeatures');
+const { validateAmplifyFlutterCoreLibraryDependency } = require('../utils/validateAmplifyFlutterCoreLibraryDependency');
 
 const platformToLanguageMap = {
   android: 'java',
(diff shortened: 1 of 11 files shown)


ragingsquirrel3:
I can verify this is working as expected in amplify-flutter with the last commit here reverted. https://github.com/aws-amplify/amplify-codegen/pull/382/commits/dd234067b4984b68c0cc209f4188fe2041af2de3 was causing error for me locally.

ragingsquirrel3:
Also, here is docs PR in draft state, would not be merged until amplify-flutter 0.4.0 is out https://github.com/aws-amplify/docs/pull/3979

packages/amplify-codegen/src/utils/validateAmplifyFlutterCoreLibraryDependency.js
```diff
@@ -0,0 +1,9 @@
+const { validateAmplifyFlutterVersion } = require('./validateAmplifyFlutterVersion');
+
+const MINIMUM_VERSION_CONSTRAINT = '>= 0.4.0 || >= 0.4.0-rc.2';
```
ragingsquirrel3, line 3: nit/question: Could the rc version be 0.4.0-rc.1 instead of 2? Unless I'm missing some reason for wanting 2.

packages/appsync-modelgen-plugin/src/visitors/appsync-dart-visitor.ts
```diff
@@ -361,20 +368,28 @@ export class AppSyncModelDartVisitor<
     }
     //other getters
     if (this.isNullSafety()) {
+      let forceCastException = `throw new DataStoreException(
+      DataStoreExceptionMessages.codeGenRequiredFieldForceCastExceptionMessage,
+      recoverySuggestion:
+        DataStoreExceptionMessages.codeGenRequiredFieldForceCastRecoverySuggestion,
+      underlyingException: e.toString()
+      );`;
+      if (this.config.dartUpdateAmplifyCoreDependency === true) {
```
AaronZyLee, line 384: Just want to confirm if this exception change is expected along with flutter zero four. Otherwise LGTM 🥇

AaronZyLee (approved):

packages/amplify-codegen/src/commands/models.js
```diff
@@ -107,6 +109,8 @@ async function generateModels(context) {
     // override isTimestampFieldsAdded to true when using amplify-flutter > 0.3.0 || > 0.3.0-rc.2
     isTimestampFieldsAdded = validateAmplifyFlutterCapableZeroThreeFeatures(projectRoot);
     enableDartZeroThreeFeatures = validateAmplifyFlutterCapableZeroThreeFeatures(projectRoot);
+    // This feature is supported only for users using amplify-flutter > 0.4.0 || > 0.4.0-rc.2
```
phani-srikar, line 112: nit: change this to `rc.1`

packages/appsync-modelgen-plugin/src/visitors/appsync-dart-visitor.ts
phani-srikar: Yes these exceptions are different in `amplify_core` package. @ragingsquirrel3 can confirm.

packages/amplify-codegen/src/commands/models.js
dpilch: Fixed

PR #398, opened 2022-03-10
fix(appsync-modelgen-plugin): removes hardcoding of syncable flag

### LINKED ISSUE #397: `amplify codegen models` generates model types marked as syncable when they shouldn't be
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-codegen/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-codegen/blob/master/CONTRIBUTING.md#bugs).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

npm -g i @aws-amplify/cli

### If applicable, what version of Node.js are you using?

14.19.0

### Amplify CLI Version

7.6.23

### What operating system are you using?

Mac

### Amplify Codegen Command

codegen models

### Describe the bug

When a GraphQL schema contains an Amplify directive such as the following:

```graphql
type Comment @model(subscriptions: null) {
  id: ID!
  content: String
  todo: Todo @connection(name: "TodoComments")
}
```

(Taken from the official docs [here](https://docs.amplify.aws/cli-legacy/graphql-transformer/examples/#filter-subscriptions-by-model-fields-andor-relations))

This prevents Amplify from generating create, update and delete subscriptions, but it doesn't prevent Amplify DataStore from attempting to subscribe to those non-existent subscriptions. In order to do that the `syncable` field on the generated should be set to false.

Others have already had to jerry-rig a solution, and I confirmed it does work, however we should really fix the root cause, which I believe is [here](https://github.com/aws-amplify/amplify-codegen/blob/master/packages/appsync-modelgen-plugin/src/visitors/appsync-json-metadata-visitor.ts#L196).

### Expected behavior

`amplify codegen models` should generate a `schema.js` with `syncable: false` for all models with a `@model(subscriptions: null)` directive.

### Reproduction steps

1. Create a GraphQL schema such as the above
2. Run `amplify codegen models`
3. Inspect `schema.js` and notice that the `Comment` model type has a `syncable` value of `true` not `false`.

### GraphQL schema(s)

_No response_

### Log output

_No response_

### Additional information

This bug has been around a long time and has been reported in other issues such as:

* aws-amplify/amplify-js#6260
  * @TheMoums even created a [workaround](https://github.com/aws-amplify/amplify-js/issues/6260#issuecomment-831080188) to monkey-patch the generated `schema.js` in order to get something useable in DataStore.
* aws-amplify/amplify-cli/issues/3114
  * A DataStore error which is a direct result of the incorrect `syncable` value being generated by `amplify codegen models`

danrivett:
Fixes #397.

#### Description of changes
Updates appsync-json-metadata-visitor to not always hardcode 'syncable: true' for model definitions,
but instead checks if the model directive has 'subscriptions: null' defined.

If so, sets 'syncable: false' since no AppSync Subscriptions are generated for that model.

#### Issue #, if available
#397

Ran unit tests and ran `amplify codegen models` locally and verified result.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-json-metadata-visitor.test.ts
+++ b/packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-json-metadata-visitor.test.ts
@@ -27,7 +27,6 @@ const getVisitor = (schema: string, target: 'typescript' | 'javascript' | 'typeD
 };
 
 describe('Metadata visitor', () => {
-
   const schema = /* GraphQL */ `
     type SimpleModel @model {
       id: ID!
@@ -42,6 +41,18 @@ describe('Metadata visitor', () => {
     type SimpleNonModelType {
       id: ID!
       names: [String]
+    }
+
+    type ModelWithoutSubscriptions @model(subscriptions: null) {
+      id: ID!
+      name: String
+      bar: String
+    }
+
+    type ModelWithCustomSubscriptions @model(subscriptions: ["onCustomUpdate"]) {
+      id: ID!
+      name: String
+      bar: String
     }
   `;
   let visitor: AppSyncJSONVisitor;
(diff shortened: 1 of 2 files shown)


AaronZyLee (approved):
LGTM 👍

manueliglesias:
`syncable: false` would cause the model to not be syncable at all (**including queries and mutations**)

There’s a difference from saying _"I don’t want real time data from this model"_ and _"I don’t want this model to sync with a backend"_

danrivett:
Good point @manueliglesias. The thing that clouds this is that Amplify DataStore (which we're using) relies heavily on subscriptions and errors on start if `@model(subscriptions: null)` is set when `syncable` is hard-coded to true like it currently is.

I believed 'syncable' was analogous to where 'realtime' updates should occur - e.g. should subscriptions be created; rather than have any effet on manual queries/mutations.

Currently the model types we have `syncable: false` we can still query and mutate those models on the backend, but we aren't using DataStore there.

Either way, it seems like there is a gap here where if a model is marked as `@model(subscriptions: null)` then DataStore should not attempt to create subscriptions and sync data - which was what I thought the `syncable` flag was designed for. If that's not the case perhaps there's a better approach to avoid errors.

manueliglesias:
Hey @danrivett 

>Either way, it seems like there is a gap here where if a model is marked as @model(subscriptions: null) then DataStore should not attempt to create subscriptions and sync data - which was what I thought the syncable flag was designed for. If that's not the case perhaps there's a better approach to avoid errors.

Yeah, this is something we are aware of and thinking about solutions. We want to support having `syncable` models that don't have real time updates. In more practical terms this might mean [adding a condition in the `forEach` inside the `start` function of the subscriptions processor](https://github.com/aws-amplify/amplify-js/blob/348366a044be2a3364c956f3a59ea125a7fb7d58/packages/datastore/src/sync/processors/subscription.ts#L302-L305) to inspect the model definition and skip the generation of subscriptions for the ones that have `subscriptions: null` (PRs welcomed if you want to take a crack at it)

manueliglesias:
hey @AaronZyLee , is the `subscriptions: null` piece somehow preserved and added to the `schema.js`?

AaronZyLee:
> hey @AaronZyLee , is the `subscriptions: null` piece somehow preserved and added to the `schema.js`?

@manueliglesias We do not process that piece in current modelgen or preserve it either. It is only being processed on the transformer side.

manueliglesias:
Thanks @AaronZyLee 

@danrivett looks like the PR in DataStore would require that codegen preserves that information I talked about first.

mikepschneider:
Closing this per Manuel's request

PR #413, opened 2022-04-13
chore: remove deprecated feature flags

AaronZyLee:
Doc PR: https://github.com/aws-amplify/docs/pull/4206

#### Description of changes

The following feature flags are deprecated and are removed from codegen implementation:

- [`useAppsyncModelgenPlugin`](https://docs.amplify.aws/cli/reference/feature-flags/#useAppsyncModelgenPlugin)
- [`useDocsGeneratorPlugin`](https://docs.amplify.aws/cli/reference/feature-flags/#useDocsGeneratorPlugin)
- [`useTypesGeneratorPlugin`](https://docs.amplify.aws/cli/reference/feature-flags/#useTypesGeneratorPlugin)
- [`cleanGeneratedModelsDirectory`](https://docs.amplify.aws/cli/reference/feature-flags/#cleanGeneratedModelsDirectory)
- [`retainCaseStyle`](https://docs.amplify.aws/cli/reference/feature-flags/#retainCaseStyle)

- `yarn test`
- `amplify-dev codegen`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


david-mcafee (commented):
Can you provide a little more context around these changes? For example, how we communicated the deprecation of these feature flags? Additionally, will anything not be backwards compatible / will this be a breaking change? The process for feature flag deprecation is unclear to me ATM. Thank you!!

AaronZyLee:
> Can you provide a little more context around these changes? For example, how we communicated the deprecation of these feature flags? Additionally, will anything not be backwards compatible / will this be a breaking change? The process for feature flag deprecation is unclear to me ATM. Thank you!!

The first three feature flags were added at the time when we migrated codegen plugin packages(modelgen, typegen and docgen) to the new repo and renamed them. As the time elapsed, we continued to update the new packages and remained old unchanged. The migration is completed and new packages are now widely used. On the other hand, some old projects still use the deprecated packages which will introduce some problems(as suggested by some recent issues). Therefore, these feature flags should be removed to avoid these situations.

For `cleanGeneratedModelDirectory`, this was reverted previously and never in use. So it is removed as well.

For `retainCaseStyle`, this is for naming convention of generating graphql operations in doc-gen. This was introduced at the same time as flags above and is safe to remove.

For the communication, I asked in the CLI slack channel and still have no response yet. I looked into the feature flag code base and did not find any related part regarding sending messages(in console when running commands) for deprecated ff. Personally I think this part should stay in the CLI feature flag implementations rather than the codegen package. So in my PR I just remove the usage in codegen. Additionally, I have a drafted PR in doc repo which serves as a communication to the customer.

phani-srikar (approved):
The changes look good to me. Is there a reason we are not removing the other deprecated FFs like [addTimestampFields](https://docs.amplify.aws/cli/reference/feature-flags/#addTimestampFields) etc?

AaronZyLee:
> The changes look good to me. Is there a reason we are not removing the other deprecated FFs like [addTimestampFields](https://docs.amplify.aws/cli/reference/feature-flags/#addTimestampFields) etc? This should probably be accompanied with changes to CLI to throw the warning if customers are using non-default value of deprecated FFs.

The feature flag will be removed in order. This PR removes those earliest ones and the rest will be removed in the incoming patch version.

I agree that we need a warning message mechanism implemented for feature flags. This will need coordination with CLI team since it is a generic approach for all feature flags rather than codegen itself.

PR #417, opened 2022-04-15
build: fix alert for lodash version under 4.17.21

AaronZyLee:
#### Description of changes

Bump dependencies to the version where the `lodash` is greater than `4.17.21` to fix the dependent bot alert and avoid injection attack

Examine the `yarn.lock` file for package `lodash` to ensure that all resolved versions are greater than `4.17.21`

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-codegen/package.json
+++ b/packages/amplify-codegen/package.json
@@ -23,7 +23,7 @@
     "@aws-amplify/appsync-modelgen-plugin": "1.32.0",
     "@aws-amplify/graphql-docs-generator": "2.4.2",
     "@aws-amplify/graphql-types-generator": "2.8.6",
-    "@graphql-codegen/core": "1.8.3",
+    "@graphql-codegen/core": "2.5.1",
     "chalk": "^3.0.0",
     "fs-extra": "^8.1.0",
     "glob-all": "^3.1.0",
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "@typescript-eslint/parser": "^2.16.0",
     "cmd-shim": "^3.0.3",
     "codecov": "^3.7.0",
-    "commitizen": "^3.1.2",
+    "commitizen": "^4.2.4",
     "copyfiles": "^2.2.0",
     "cz-conventional-changelog": "^3.0.2",
     "eslint": "^6.8.0",
--- a/packages/amplify-codegen-e2e-tests/package.json
+++ b/packages/amplify-codegen-e2e-tests/package.json
@@ -23,7 +23,7 @@
   },
   "dependencies": {
     "amplify-codegen-e2e-core": "1.1.6",
-    "aws-amplify": "^3.0.8",
+    "aws-amplify": "^4.3.20",
     "aws-appsync": "^4.0.3",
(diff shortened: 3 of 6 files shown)


packages/amplify-codegen/package.json
```diff
@@ -23,7 +23,7 @@
     "@aws-amplify/appsync-modelgen-plugin": "1.32.0",
     "@aws-amplify/graphql-docs-generator": "2.4.2",
     "@aws-amplify/graphql-types-generator": "2.8.6",
-    "@graphql-codegen/core": "1.8.3",
+    "@graphql-codegen/core": "2.5.1",
```
alharris-at, line 26: I think the graphql-codegen/core dep bump might actually be what caused some issues in studio testing a few weeks back. Is this required for the upgrade?

sachscode: @alharris-at  The older version of graphql-codegen/core pull in an older version of lo-dash which has two dependabot warnings 
1.  high          │ Command Injection in lodash  │ Patched in    │ >=4.17.21                                                    │

    │ @aws-amplify/cli-internal > amplify-container-hosting >      │
    │ @aws-amplify/amplify-category-api >                          │
    │ amplify-provider-awscloudformation > amplify-codegen >       │
    │ @graphql-codegen/core > @graphql-codegen/plugin-helpers >    │
    │ @graphql-toolkit/common > lodash                             │
 More info     │ https://www.npmjs.com/advisories/1068132

alharris-at: We had to revert this PR previously https://github.com/aws-amplify/amplify-codegen/pull/325/files due to studio impact. Are we able to just bump the lodash dep as an override or something along those lines?

alharris-at: I suspect if we move forward with this change we'll fail studio smoke tests during the CLI release process, so just trying to get ahead of that.

dpilch: What is the minimum version of lodash needed to fix the vulnerability?

AaronZyLee: The mininum version is 4.17.21. We do not have direct dependency but a downstream dependency of `graphql-codegen` packages

dpilch: I tried changing `@graphql-codegen/core` to `1.17.10` to avoid the breaking change. It appears that this version fixes the lodash dependency issue. 

```
> yarn list lodash 
yarn list v1.22.18
warning Filtering by arguments is deprecated. Please use the pattern option instead.
└─ lodash@4.17.21
```

AaronZyLee: That might fix but still we do not know the impact of the version `1.17.10` to studio. Do we have any context from last failure for studio? We can analyze the root cause and figure out a solution to avoid the failure.

sachscode: @AaronZyLee could we please test this with Studio E2E in a tagged release.

AaronZyLee: @sachscode Sure. I will rebase this PR to a tag release branch and run codegen tag release from there.

PR #429, opened 2022-05-18
fix(appsync-dart-visitor): generate correct model id filed name

HuiSF:
#### Description of changes

Change Flutter model id field name from `<modelName>.id` to just `id`.

The original behavior should be redundant and inaccurate.

1. In the [iOS implementation of amplify-flutter](https://github.com/aws-amplify/amplify-flutter/blob/5c6d8eab63c1ea11e63d1ec8e82a975a4dffeb28/packages/amplify_datastore/ios/Classes/types/query/QueryPredicateBuilder.swift#L23-L29), it removes `<modelName>` elaborately when using the field name
2. The original name breaks query predicate functionality in amplify-flutter Android
    - base sync filter generated wrongly (see [this issue](https://github.com/aws-amplify/amplify-flutter/issues/1565))
    - `DataStore.query` was luckily working as amplify-android generated SQL commands somehow compatible with `<modelName>.id`

**NOTE:** this change won't be a breaking change, as amplify-flutter is backwards compatible with previously generated model. There is no frictions when customer upgrade either amplify-flutter library or amplify codegen.

#### Issue #, if available

https://github.com/aws-amplify/amplify-flutter/issues/1565

1. Unit tests in amplify-codegen
4. Integration tess in amplify-flutter

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-dart-visitor.test.ts.snap
+++ b/packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-dart-visitor.test.ts.snap
@@ -157,7 +157,7 @@ class Post extends Model {
     'id': id, 'title': _title, 'rating': _rating, 'status': enumToString(_status), 'comments': _comments?.map((Comment? e) => e?.toJson()).toList()
   };
 
-  static final QueryField ID = QueryField(fieldName: \\"post.id\\");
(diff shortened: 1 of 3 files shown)


dnys1 (approved):

AaronZyLee (approved):
LGTM

HuiSF:
Please go ahead merge, if no objections. Thanks!

PR #440, opened 2022-06-14
feat: add support for sub::username identity claim

danielleadams:
#### Description of changes
Related to changing the default identity claim here: https://github.com/aws-amplify/amplify-category-api/commit/fd2e4ff706ea3dccc609762a295bbeec3da20bd5. This change will take the `sub::username` identity claim and use `cognito:username` value from the token. This is primarily to account for the code generated through codegen. The libraries don't "know" how to handle a double identity claim yet (ie. sub + username), so the generated code will use the username identity claim.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/appsync-modelgen-plugin/src/utils/process-auth.ts
+++ b/packages/appsync-modelgen-plugin/src/utils/process-auth.ts
@@ -26,7 +26,7 @@ export enum AuthModelMutation {
 }
 
 const DEFAULT_GROUP_CLAIM = 'cognito:groups';
-const DEFAULT_IDENTITY_CLAIM = 'username';
+const DEFAULT_IDENTITY_CLAIM = 'sub::username';
 const DEFAULT_OPERATIONS = [AuthModelOperation.create, AuthModelOperation.update, AuthModelOperation.delete, AuthModelOperation.read];
 const DEFAULT_AUTH_PROVIDER = AuthProvider.userPools;
 const DEFAULT_OWNER_FIELD = 'owner';
@@ -51,6 +51,12 @@ export type AuthDirective = CodeGenDirective & {
   };
 };
 
+export const createIdentityClaim = (identityClaim: string): string => {
+  if (identityClaim === 'username' || identityClaim === 'sub::username') identityClaim = 'cognito:username';
+
+  return identityClaim;
+};
+
 export function processAuthDirective(directives: CodeGenDirectives): AuthDirective[] {
   const authDirectives = directives.filter(d => d.name === 'auth');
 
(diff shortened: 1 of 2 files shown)


packages/appsync-modelgen-plugin/src/utils/process-auth.ts
```diff
@@ -51,6 +51,12 @@ export type AuthDirective = CodeGenDirective & {
   };
 };
 
+export const createIdentityClaim = (identityClaim: string): string => {
+  if (identityClaim === 'username' || identityClaim === 'sub::username') identityClaim = 'cognito:username';
```
lawmicha, line 55: Codegen provides schema info for the libraries to use. There's both codegen specific logic that maps the schema's identityClaim to the codegenerated schema and then there's library logic that takes "cognito:username" and retrieves the "username" from the JWT access token's username. Since the new and future default identityClaim is "sub::username" and it produces backwards compatible APIs (subscription operations have `owner` input with the value being JWT access token's username value) then "cognito:username" as the output is the correct value for the libraries to continue to mapping "cognito:username" to the JWT accessToken.username.

phani-srikar (approved):
LGTM. Better to test this e2e in coordination with frontend libraries. Since there are no changes to tests that check the `identityClaim` value, it is being continued to be mapped to `cognito:username`.

packages/appsync-modelgen-plugin/src/utils/process-auth.ts
lawmicha: https://github.com/aws-amplify/amplify-android/blob/main/core/src/main/java/com/amplifyframework/core/model/AuthRule.java#L125-L134

>  An older version of the CLI incorrectly generated a value of "cognito:username" so we also check for this value and convert it to the proper default of "username" for backwards compatibility.

lawmicha: So I think i'm coming to the conclusion that if we are introducing a new identityClaim who's value from the JWT token comes from "username" then "sub::username" (public facing) should return "username" (codegen library). We might as well fix the problem for username as well, by returning
```
// Ideal
  if identityClaim === 'sub::username' 
     identityClaim = 'username';
  // if "identityClaim == 'username' // do nothing, written here to show that it's a 1:1 mapping
     identityClaim == 'username'
``

lawmicha (changes requested):
I think, public facing "sub::username" should codegen "username". We should do some due dilligence around this by testing library's usage, to make sure there's no library code that does a mapping for "username", but i suspect library code uses "username" as is.

"sub:username" codegenerating "cognito:username" does not seem like the correct thing to do for a new identityClaim type, even though we have a permanent workaround in the library code that maps "cognito:username" to "username" (for backwards compat of older codegenerated models containing "cognito:username")

danielleadams:
@lawmicha should we do some work to convert `cognito:username` -> `username`? Codegen changes `username` -> `cognito:username`, and then `sub::username` uses `username`, so that logic is a little confusing (as implemented). I do agree that everything should just point to `username`.

lawmicha:
> @lawmicha should we do some work to convert `cognito:username` -> `username`? Codegen changes `username` -> `cognito:username`, and then `sub::username` uses `username`, so that logic is a little confusing (as implemented). I do agree that everything should just point to `username`.

I think we should work backwards from what is the information we are placing in the code generated file? I think the codegenerated file should explicitly tell the library which part of the claim to access out of the token. So, codegen process owns the logic for translating the source to the result being "username". Otherwise, if the information is just passthrough, then codegen process should be simple, have no internal translation, and represent schema information directly in the codegenerated file. @phani-srikar  do you have any thoughts on this?

danielleadams:
I'm going to close this since it's gotten stale. I will create a backlog item to address this with codegen.

PR #475, opened 2022-08-24
feat: add lazy loading types [WIP]

dpilch:
#### Description of changes

* Add lazy loading types to modelgen
* Add feature flag `useLazyLoading` to enable use of lazy loading
* 
Todo:
* Other platforms (only js is done)
* Add versioning schema json as runtime feature flag

#### Issue #, if available

N/A

Unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-codegen/src/commands/models.js
+++ b/packages/amplify-codegen/src/commands/models.js
@@ -82,9 +82,10 @@ async function generateModels(context) {
 
   const generateIndexRules = readFeatureFlag('codegen.generateIndexRules');
   const emitAuthProvider = readFeatureFlag('codegen.emitAuthProvider');
-  const usePipelinedTransformer = readFeatureFlag('graphQLTransformer.useExperimentalPipelinedTransformer')
+  const usePipelinedTransformer = readFeatureFlag('graphQLTransformer.useExperimentalPipelinedTransformer');
   const transformerVersion = readNumericFeatureFlag('graphQLTransformer.transformerVersion');
   const respectPrimaryKeyAttributesOnConnectionField = readFeatureFlag('graphQLTransformer.respectPrimaryKeyAttributesOnConnectionField');
+  const useLazyLoading = readFeatureFlag('graphQLTransformer.useLazyLoading');
 
   let isTimestampFieldsAdded = readFeatureFlag('codegen.addTimestampFields');
   let enableDartNullSafety = readFeatureFlag('codegen.enableDartNullSafety');
@@ -128,6 +129,7 @@ async function generateModels(context) {
       transformerVersion,
       dartUpdateAmplifyCoreDependency,
       respectPrimaryKeyAttributesOnConnectionField,
+      useLazyLoading,
     },
   });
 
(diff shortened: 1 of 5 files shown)


packages/amplify-codegen/src/commands/models.js
```diff
@@ -44,6 +45,85 @@ const readNumericFeatureFlag = key => {
...
 
+/**
+ * Returns the version number of package installed
+ * @param {string} target language target
+ * @returns
+ */
+const readDependencyVersionInstalled = (dependency, target) => {
+  // TODO: other platforms and error handling
+  const projectPath = pathManager.findProjectRoot();
+  switch (target) {
```
phani-srikar, line 56: If we're going ahead with this, it has to be consistent across platforms. We should be able to answer the question, is this possible for native platforms?

packages/amplify-codegen/src/commands/models.js
```diff
@@ -44,6 +45,85 @@ const readNumericFeatureFlag = key => {
...
+    case 'swift':
+      return '';
+    case 'dart':
+      return '';
+    case 'javascript':
+      // there are a few ways to get the version of an installed package
+      // just choosing one for PoC
+      // Will assess the best option later
+      const dependencyInstallPath = path.join(projectPath, 'node_modules', dependency, 'package.json');
+      if (fs.pathExistsSync(dependencyInstallPath)) {
```
phani-srikar, line 68: I think we need to check for the lock file rather than the package.json file itself.

packages/amplify-codegen/src/commands/models.js
```diff
@@ -44,6 +45,85 @@ const readNumericFeatureFlag = key => {
...
+    dart: '',
+    javascrpt: '5.0.0',
+  };
+  const minLanguageDataStoreVersionMap = {
+    java: '',
+    swift: '',
+    dart: '',
+    javascrpt: '4.0.0',
+  };
+  const lazyLoadingFeatureFlag = readFeatureFlag('graphQLTransformer.useLazyLoading');
```
phani-srikar, line 91: Why are we combining lib version check with FF?

packages/amplify-codegen/src/commands/models.js
```diff
@@ -44,6 +45,85 @@ const readNumericFeatureFlag = key => {
...
+      const minVersionString = `${dependencyInstalled}@${
+        onlyDataStoreInstalled ? minLanguageDataStoreVersionMap[target] : minLanguageLibraryVersionMap[target]
+      }`;
+      context.print.warning(
+        `"useLazyLoading" feature flag is enabled, but the installed library version (${libraryVersionString}) does not support lazy loading. Models will be generated with lazy loading disabled. Library version must be at least ${minVersionString} to enable lazy loading.`,
+      );
+      return false;
+    }
+  } catch {
+    context.print.warning('Unable to infer lazy loading based on Amplify library version. Using "useLazyLoading" feature flag instead.');
```
phani-srikar, line 122: This needs a review before we print this out for customers

phani-srikar (changes requested):

PR #476, opened 2022-08-24
feat(amplify-codegen): new codegen command of  model intropection

AaronZyLee:
#### Description of changes

Add new codegen command which will generate the `model-introspection.json` file under the given directory path.
```bash
$ amplify codegen model-introspection --output-dir ${dir_path_input}
```
The changes include:
- New introspection visitor
- JSON schema and its validator
- Script for generating JSON schema from TS type definition file
- Unit and E2E tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-codegen/commands/codegen/model-introspection.js
+++ b/packages/amplify-codegen/commands/codegen/model-introspection.js
@@ -1,19 +1,12 @@
 const codeGen = require('../../src');
 const { exitOnNextTick } = require('amplify-cli-core');
 const featureName = 'model-introspection';
-const path = require('path');
 
 module.exports = {
   name: featureName,
   run: async context => {
     try {
-      // Verify override path flag is provided
-      const outputDirParam = context.parameters.options ? context.parameters.options['output-dir'] : null;
-      if ( !outputDirParam ) {
-        throw new Error('Expected --output-dir flag to be set for model introspection command.');
-      }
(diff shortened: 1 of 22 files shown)


packages/amplify-codegen/commands/codegen/model-introspection.js
```diff
@@ -0,0 +1,24 @@
...
+const path = require('path');
+
+module.exports = {
+  name: featureName,
+  run: async context => {
+    try {
+      // Verify override path flag is provided
+      const outputDirParam = context.parameters.options ? context.parameters.options['output-dir'] : null;
+      if ( !outputDirParam ) {
+        throw new Error('Expected --output-dir flag to be set for model introspection command.');
```
alharris-at, line 13: Do we want this to be an output dir, or an output path for the specific file?

alharris-at (commented):
Overall the change LGTM - using TS to generate the json schema should make the process significantly more readable than hand-writing it. Only comment is around whether we want the output-dir or an output-path instead.

packages/amplify-codegen/commands/codegen/model-introspection.js
AaronZyLee: Confirmed with studio team. They agree with the output directory option with the hard-coded file name `model-introspection.json`. The name does not matter but it should keep unchanged.

alharris-at (approved):

PR #478, opened 2022-08-31
Add support for output dir arg on models

marcvberg:
#### Description of changes

Add support for new `--output-dir` flag when invoking models directly. The parameter is not required

#### Issue #, if available

N/A

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-codegen/src/commands/models.js
+++ b/packages/amplify-codegen/src/commands/models.js
@@ -78,7 +78,7 @@ async function generateModels(context, outputDirPath = null, isIntrospection = f
   const schemaContent = loadSchema(apiResourcePath);
 
   const outputDirParam = context.parameters?.options?.['output-dir'];
-  if ( !outputDirPath && typeof(outputDirParam) !== 'string' ) {
+  if ( !outputDirPath && outputDirParam && typeof(outputDirParam) !== 'string' ) {
     throw new Error('Expected provided --output-dir flag to be given output location as input.');
   }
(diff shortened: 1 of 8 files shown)


packages/amplify-codegen/src/commands/models.js
```diff
@@ -76,7 +76,12 @@ async function generateModels(context, outputDirPath = null, isIntrospection = f
   });
 
   const schemaContent = loadSchema(apiResourcePath);
-  const outputPath = outputDirPath || path.join(projectRoot, getModelOutputPath(context));
+
+  const outputDirParam = context.parameters?.options?.['output-dir'];
+  if ( !outputDirPath && outputDirParam && typeof(outputDirParam) !== 'string' ) {
+    throw new Error('Expected provided --output-dir flag to be given output location as input.');
+  }
```
alharris-at, line 83: @AaronZyLee - do you think we need this validation here? This seems like it'd still work if we just yanked lines 81-83

alharris-at: Actually, I suppose it doesn't hurt anything in particular.

alharris-at (approved):

packages/amplify-codegen/src/commands/models.js
AaronZyLee: I think the validation is for the case where we provide `--output-dir` without a value, which is set as `true`. The following command should be invalid.
```
$ amplify codegen models --output-dir
```

AaronZyLee (approved):

PR #489, opened 2022-09-14
feat: lazy loading conditional type

dpilch:
#### Description of changes

This an alternative approach to https://github.com/aws-amplify/amplify-codegen/pull/475.

* Add lazy loading types in a way that is backwards compatible.
* Use [conditional types](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html) to use lazy or eager types.
* Add `pragma` number to schema JSON file to be used by library.
  * Using version number from `package.json`,

The following types will be exported from the DataStore library. 

```
export declare class Enabled {
  enabled: true;
}
export declare class Disabled {
  disabled: true;
}
export declare class LazyLoading extends Enabled {}
```

Then imported in the generated type declaration file.

Normally when importing a non-existent property the build would fail. However, when doing this in a `.d.ts` file it appears to be importing it as `any`.

I tested with all major versions of TypeScript and it works with 2.x, 3.x, and 4.x. This is all but 1.x, but the issue appeared to be an incompatibility between webpack and TypeScript. Given that TypeScript 1.x is from 7 years ago it is probably not a concern.

I also tested that this works with `strict` set to `true`.

---

##### Example (old version of library):

In the case that an older version of the library is used, the `Enabled`, `Disabled`, and `LazyLoading` types won't be exported. 

```
export declare type Blog = LazyLoading extends Disabled ? EagerBlog : LazyBlog;
```

Will evaluate to:

```
export declare type Blog = any extends any ? EagerBlog : LazyBlog;
```

Giving the final type:

```
export declare type EagerBlog;
```

---

##### Example (new version of library):

```
export declare type Blog = LazyLoading extends Disabled ? EagerBlog : LazyBlog;
```

Gives the final type:

```
export declare type Blog = LazyBlog;
```

Todo:
- [x] Test linter
- [x] Test build systems
- [x] Test with all supported versions of TS
- [x] Conditionally import `AsyncCollection`, `LazyLoading`, `Enabled`, and `Disabled` only when connections are used.

#### Issue #, if available

N/A

* Unit tests
* Test new types in apps with and without DataStore library changes.

Tested with tsconfig.json

```json
{
  "compilerOptions": {
    "target": "es5",
    "lib": [
      "dom",
      "dom.iterable",
      "esnext"
    ],
    "allowJs": true,
    "skipLibCheck": true,
    "esModuleInterop": true,
    "allowSyntheticDefaultImports": true,
    "strict": true,
    "forceConsistentCasingInFileNames": true,
    "noFallthroughCasesInSwitch": true,
    "module": "esnext",
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "isolatedModules": true,
    "noEmit": true,
    "jsx": "react-jsx"
  },
  "include": [
    "src"
  ]
}
```

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-javascript-visitor.test.ts.snap
```diff
@@ -28,7 +28,7 @@ export declare class Comment {
   };
   readonly customPostId: string;
   readonly content: string;
-  readonly post?: Post | null;
+  readonly post?: (LazyLoading extends Disabled ? Post : Promise<Post>) | null;
```
svidgen, line 31: The return types need to be optional only if Lazy loading is disabled. So, I think we need to change this to something like:

```
  readonly post: LazyLoading extends Disabled ? (Post | null | undefined) : Promise<Post | undefined>;
```

Basically, the field should not appear optional if lazy loading is enabled, but the return type from the promise could possibly be missing.

@manueliglesias, correct me if I'm wrong, but I think we also decided these promises should resolve to `undefined` instead of `null` for missing values. Is that right?

(I've left `null` in the eager version for consistency with the old types.)

packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-javascript-visitor.test.ts.snap
```diff
@@ -52,7 +52,7 @@ export declare class Post {
   };
   readonly id: string;
   readonly title: string;
-  readonly comments?: (Comment | null)[] | null;
+  readonly comments?: (LazyLoading extends Disabled ? (Comment | null)[] : AsyncCollection<Comment>) | null;
```
svidgen, line 55: Similarly to the above, I think we want to move the optionality into the type unions so that the collection is always present -- so that the developer doesn't have to drop `?`'s and/or `!`'s everywhere.

We had also decided that due to the nature of a has-many, we should always return an `AsyncCollection` for Lazy Loading, but that it may simply be empty. So, I think the backwards compatible version needs to look something like this:

```
  readonly comments: LazyLoading extends Disabled ? ((Comment | null)[] | undefined) : AsyncCollection<Comment>;
```

svidgen (commented):
Left a few comments on the specific field definitions to better align with how optional fields should be handled.

Also, do we need test with any specific build systems? Or linters or linting rules?

packages/appsync-modelgen-plugin/tsconfig.json
```diff
@@ -1,13 +1,18 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "rootDir": "src",
-    "outDir": "lib"
+    "rootDir": ".",
```
dpilch, line 4: These changes, along with the addition of `src/tsconfig.json` are necessary for importing the version from `package.json`.

packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-json-metadata-visitor.test.ts.snap
```diff
@@ -317,6 +317,7 @@ exports[`Metadata visitor for custom PK support relation metadata for hasMany &
     },
     \\"enums\\": {},
     \\"nonModels\\": {},
+    \\"pragma\\": \\"2.1.0\\",
```
phani-srikar, line 320: can you please tell me how this will be used?

packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-javascript-visitor.test.ts.snap
```diff
@@ -17,11 +19,29 @@ export declare class Post {
...
+  readonly title: string;
+  readonly comments: AsyncCollection<Comment>;
+  readonly createdAt?: string | null;
+  readonly updatedAt?: string | null;
+  
+}
+
+export declare type Post = LazyLoading extends Disabled ? EagerPost : LazyPost
+
+export declare const Post: (new (init: ModelInit<Post>) => Post) & Post & {
```
phani-srikar, line 40: slightly confused with this export here. Are we ultimately just exporting a type instead of a class?

phani-srikar (changes requested):
Looks good overall. I have got a couple of questions:
- Will the old version of codegen be compatible to the new version of library?
- others in the comments

packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-json-metadata-visitor.test.ts.snap
dpilch: The new version of the library will not be compatible with the old version of codegen. This will be used for the library to determine what version of codegen was used. In the case of the new version of the library with the old version of codegen there will be an error along the lines of `Codegen version x.x.x is not supported please use ^x.x.x`.

dpilch: Naming isn't set in stone. `version` was already taken, but we could do something like `codegenVersion`.

packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-javascript-visitor.test.ts.snap
dpilch: Previously the the type declaration used a `class` which can be used as a `value` and a `type` in TypeScript. It was not possible to do conditional typing with a class, so we needed to use this technique ([Declaration Merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html)).


Without exporting both `type` and `const` here we would get errors:

* without `type`: `'Post' refers to a value, but is being used as a type here. Did you mean 'typeof Post'`
  * We could have customers fix this themselves when upgrading to version 5, but we decided this would be too much friction.
* without `const`: `'Post' only refers to a type, but is being used as a value here.`

packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-json-metadata-visitor.test.ts.snap
AaronZyLee: Some thoughts on the `pragma`:  Although the `appsync-modelgen-plugin` package version has direct relation with the library, as a CLI user it is not easy to directly consume this package. The current plugin relation is `amplify-cli => amplify-codegen => appsync-modelgen-plugin`.  My suggestion is to use the version of `amplify-codegen` for this value as it is within the same mono-repo structure and have fixed dependency of `appsync-modelgen-plugin`

AaronZyLee: Alternatively we can figure out a way to determine the specific `amplify-codegen` version given the existing `appsync-modelgen-plugin` package

dpilch: Makes sense, I think it could be passed as config here then: https://github.com/aws-amplify/amplify-codegen/blob/main/packages/amplify-codegen/src/commands/models.js#L119.

dpilch:
Another thing I forgot to mention, the type flag names are still up in the air as well (`LazyLoading`, `Enabled`, and `Disabled`). Another option could be `V5Model` so that it is not directly linked to lazy loading.

dpilch:
I tested a few build systems and they all seem fine. They all appear to just use the TypeScript  compiler (which makes sense). So as along as the TypeScript compiler is working we are good to go.

I also tested linting. It does fail the `@typescript-eslint/ban-ts-comment` rule becuase we are using `@ts-ignore`. I don't think this is a concern though because the CLI will add `src/models` to `.eslintignore`. In order to get the error I had to use `eslint . --no-ignore`.

We could add `// eslint-disable-next-line @typescript-eslint/ban-ts-comment` to the generation, but I'm not sure it is necessary.

packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-json-metadata-visitor.test.ts.snap
phani-srikar: If you're using this to store codegen version to check against in the library, I would suggest re-naming this to `codegenVersion` to make it more obvious and also use the version of the `amplify-codegen` package like Aaron mentioned.

packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-javascript-visitor.test.ts.snap
phani-srikar: Got it, so this is needed to support Declaration Merging.

phani-srikar:
> Another thing I forgot to mention, the type flag names are still up in the air as well (`LazyLoading`, `Enabled`, and `Disabled`). Another option could be `V5Model` so that it is not directly linked to lazy loading.

I believe this needs to go through API review if you haven't had one. For me, `LazyLoading` makes more sense than `V5Model` but not sure about `Enabled`, `Disabled`.

packages/appsync-modelgen-plugin/src/visitors/appsync-json-metadata-visitor.ts
```diff
@@ -150,6 +152,7 @@ export class AppSyncJSONVisitor<
       models: {},
       enums: {},
       nonModels: {},
+      codegenVersion: this._parsedConfig.codegenVersion || 'unknown',
```
phani-srikar, line 155: Do you have the `unknown` case accounted for in the JS library code?

phani-srikar (approved):

packages/appsync-modelgen-plugin/src/visitors/appsync-json-metadata-visitor.ts
```diff
@@ -107,6 +108,7 @@ export class AppSyncJSONVisitor<
   ) {
     super(schema, rawConfig, additionalConfig, defaultScalars);
     this._parsedConfig.metadataTarget = rawConfig.metadataTarget || 'javascript';
+    this._parsedConfig.codegenVersion = rawConfig.codegenVersion;
```
AaronZyLee, line 111: This line can be moved to `appsync-visitor.ts` as it is originally defined in the super class and can be consumed by other libraries later.

packages/appsync-modelgen-plugin/src/visitors/appsync-json-metadata-visitor.ts
dpilch: Yes, I have it throw an error that it cannot determine the version.

dpilch:
I've found some other errors when setting up the project in a different way. There is an easy fix, but you can hold off review until those are up.

packages/appsync-modelgen-plugin/src/visitors/appsync-json-metadata-visitor.ts
AaronZyLee: Also we need to make sure this new field not breaking the studio as the studio is in a current transient from `schema.js` to model introspection schema

AaronZyLee (changes requested):

packages/appsync-modelgen-plugin/src/visitors/appsync-json-metadata-visitor.ts
dpilch: I've talked to Chris before and he said there shouldn't be an issue, but I'll ask him to take a look at this PR.

cwoolum (commented):
Small comment

dpilch:
I've finished up the fixes necessary. I've also created a testing repo that other can use if they like. 

https://github.com/dpilch/lazy-loading-codegen

packages/appsync-modelgen-plugin/src/visitors/appsync-typescript-visitor.ts
```diff
@@ -159,29 +157,36 @@ export class AppSyncModelTypeScriptVisitor<
...
     modelObj.fields.forEach((field: CodeGenField) => {
-      modelDeclarations.addProperty(this.getFieldName(field), this.getNativeType(field), undefined, 'DEFAULT', {
+      const fieldName = this.getFieldName(field);
+      eagerModelDeclaration.addProperty(fieldName, this.getNativeType(field), undefined, 'DEFAULT', {
         readonly: true,
         optional: field.isList ? field.isListNullable : field.isNullable,
       });
+      lazyModelDeclaration.addProperty(fieldName, this.getNativeType(field, { lazy: true }), undefined, 'DEFAULT', {
+        readonly: true,
+        optional: !this.isModelType(field) && (field.isList ? field.isListNullable : field.isNullable),
```
AaronZyLee, line 188: The nullability for lazy/eager is different for model types, in which in lazy mode the model type is always optional. Is this an intentional change?

AaronZyLee (dismissed):
Overall LGTM. Just have one following question

packages/appsync-modelgen-plugin/src/visitors/appsync-typescript-visitor.ts
dpilch: It should be the model type is always _not_ optional. A model with a connection will always have a `AsyncCollection` or `AsyncItem` in the connection field. The `AsyncItem` may resolve to `undefined` and the `AsyncCollection` may resolve to an empty list to indicate the field is empty.

We are using some type mappings to allow the fields to not be supplied when creating an instance.

```
type Parent @model {
  child: Child
}
type Child @model {
  ...
}
```

```js
new Parent({});

// equivalent to

new Parent({ child: new AsyncItem((resolve) => { resolve(undefined); });
```

AaronZyLee: Ah I forget `optional: false` means always non-nullable. Make sense to me

packages/appsync-modelgen-plugin/src/visitors/appsync-typescript-visitor.ts
```diff
@@ -191,43 +196,20 @@ export class AppSyncModelTypeScriptVisitor<
...
-        },
-      ],
-      'DEFAULT',
-      {},
-    );
-
-    // copyOf method
+    const modelVariableBuilder = [
+      `export declare const ${modelName}: (new (init: ModelInit<${modelName}${modelMetaDataDeclaration}>) => ${modelName})`,
+      modelName,
```
svidgen, line 203: I think we need to remove the model name from the intersection. (Am I correctly reading that this line is responsible for that?) It causes the fields from the type to appear as static properties on the class, which I don't believe we want ... unless you found a specific case where it's needed?

svidgen (changes requested):

packages/appsync-modelgen-plugin/src/visitors/appsync-typescript-visitor.ts
dpilch: iirc this was needed for the correct typing on predicates. But I'll test it out some more.

dpilch: Predicates are working fine after removing this. Maybe it was fixed elsewhere or was never an issue ¯\_(ツ)_/¯

svidgen: Well, that's good news! That will keep the class clean at the static level. Thanks for digging on it!

AaronZyLee (approved):

PR #500, opened 2022-10-25
fix: lazy codegen for schemas without relationships

dpilch:
#### Description of changes

Always import `LazyLoading` and `LazyLoadingDisabled` in TypeScript generation. 

#### Issue #, if available

Any schema that does not contain any relationships will result in a type declaration file for TypeScript that is invalid. This is effecting >= 10.3.1 of the CLI and >= 3.2.0 of codegen.

Minimum reproduction:
```
type Todo @model {
  id: ID!
  name: String!
}
```

The generated file 
```
import { ModelInit, MutableModel } from "@aws-amplify/datastore";

type EagerTodo = {
  readonly id: string;
  readonly name: string;
}

type LazyTodo = {
  readonly id: string;
  readonly name: string;
}

export declare type Todo = LazyLoading extends LazyLoadingDisabled ? EagerTodo : LazyTodo

export declare const Todo: (new (init: ModelInit<Todo>) => Todo)
```

The file is missing the lines.
```
// @ts-ignore
import { LazyLoading, LazyLoadingDisabled } from "@aws-amplify/datastore";
```

This results in a compilation error.

```
error TS2304: Cannot find name 'LazyLoading'.
```

* Unit tests
* Create an app with min repro and attempt to build

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-javascript-visitor.test.ts
+++ b/packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-javascript-visitor.test.ts
@@ -368,6 +368,8 @@ describe('Javascript visitor with default owner auth', () => {
       validateTs(declarations);
       expect(declarations).toMatchInlineSnapshot(`
         "import { ModelInit, MutableModel } from \\"@aws-amplify/datastore\\";
+        // @ts-ignore
+        import { LazyLoading, LazyLoadingDisabled } from \\"@aws-amplify/datastore\\";
 
         export enum SimpleEnum {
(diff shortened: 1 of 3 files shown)


phani-srikar (commented):
The `options?.lazy` wasn't being set for models without relationships?

dpilch:
It wouldn't make it past line 297. `if (this.isModelType(field)) {`. In the case that there were no relationships in the schema this would never execute.

phani-srikar (approved):

svidgen (approved):
Makes sense to me. Thank-you!

PR #507, opened 2022-11-02
fix: addressing missing key for unidirectional @hasMany on JS

alharris-at:
#### Description of changes
Today, if you add a `@hasMany` directive to your schema without a corresponding `@belongsTo` on the related model, then the JS library will fail to cascade delete the related item. It appears this is because the Javascript Model Intermediate Platform Representation (MIPR) file is missing an implicit `key` in the related model.

This is being generated implicitly in the transformer today, since a DynamoDB GSI is required to query correctly, but this appears to have been missed in Codegen when the GQLv2 transformer support was launched at the end of 2021.

Because we haven't confirmed this behavior is in error on other platforms, I'm implementing this behind a boolean switch, so only the Javascript, Typescript, JSON, and Model Introspection outputs will see this change, but Flutter, Java, and Swift will not change.

#### Issue #, if available
N/A, reported internally.

The missing `key` as an issue was verified by manually editing the MIPR, and running a minimal reproduction of the issue.

Repro Schema:

```graphql
type Foo @model @auth(rules: [{ allow: public }]) {
  title: String
  bars: [Bar] @hasMany
}

type Bar @model @auth(rules: [{ allow: public }]) {
  title: String
}
```

Repro App Code:
```javascript
import { useState } from 'react';
import { DataStore, Amplify } from 'aws-amplify';
import awsconfig from "./aws-exports";
import { Foo, Bar } from './models';

Amplify.configure(awsconfig);

export default function App() {
  const [currentFoo, setCurrentFoo] = useState();

  const create = async () => {
    const foo = await DataStore.save(new Foo({ title: "foo" }));
    await DataStore.save(new Bar({ title: "bar", fooBarsId: foo.id }));
    setCurrentFoo(foo);
  };

  const deleteFoo = async () => {
    await DataStore.delete(currentFoo);
    setCurrentFoo(null);
  };

  return (
    <div>
      <button onClick={create}>Create</button>
      <button onClick={deleteFoo}>Delete</button>
    </div>
  );
}
```

Exploratory testing against this case, to ensure correctness.

Additionally, I will verify the following cases in order to protect against regressions via snapshot tests:

- [x] When parent model has implicit primary key, and there is no belongsTo directive then the implicit `key` should be added.
- [x] When parent model has explicit primary key (`id` field), and there is no belongsTo directive then the implicit `key` should be added.
- [x] When parent model has explicit primary key (custom `name` field), and there is no belongsTo directive then the implicit `key` should be added.
- [x] When parent model has explicit primary key (custom `name` field, sort key), and there is no belongsTo directive then the implicit `key` should be added.

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/appsync-modelgen-plugin/src/__tests__/utils/process-has-many.test.ts
+++ b/packages/appsync-modelgen-plugin/src/__tests__/utils/process-has-many.test.ts
@@ -246,6 +246,13 @@ describe('addHasManyKey', () => {
      *   id: ID!
      * }
      */
+    const hasManyField: CodeGenField = {
+      name: 'bar',
+      type: 'Bar',
+      isList: true,
+      isNullable: true,
+      directives: [{ name: 'hasMany', arguments: {} }],
+    };
     const fooModel: CodeGenModel = {
       name: 'Foo',
       type: 'model',
@@ -260,13 +267,7 @@ describe('addHasManyKey', () => {
           isNullable: false,
           directives: [],
         },
-        {
-          name: 'bar',
-          type: 'Bar',
-          isList: true,
-          isNullable: true,
-          directives: [{ name: 'hasMany', arguments: {} }],
-        },
+        hasManyField,
       ],
     };
     const associationField: CodeGenField = {
@@ -300,9 +301,9 @@ describe('addHasManyKey', () => {
       associatedWith: associationField,
       associatedWithFields: [associationField],
     };
-    const expectedKeyDirective = { name: 'key', arguments: { name: 'gsi-Foo.fooBarsId', fields: ['fooBarsId'] } };
+    const expectedKeyDirective = { name: 'key', arguments: { name: 'gsi-Foo.bar', fields: ['fooBarsId'] } };
 
-    addHasManyKey(fooModel, connection);
+    addHasManyKey(hasManyField, fooModel, connection);
     expect(barModel.directives).toEqual(expect.arrayContaining([expectedKeyDirective]));
   });
 
@@ -320,7 +321,14 @@ describe('addHasManyKey', () => {
      *   id: ID!
      * }
      */
-     const fooModel: CodeGenModel = {
+    const hasManyField: CodeGenField = {
+      name: 'bar',
+      type: 'Bar',
+      isList: true,
+      isNullable: true,
+      directives: [{ name: 'hasMany', arguments: {} }],
+    };
+    const fooModel: CodeGenModel = {
       name: 'Foo',
       type: 'model',
       directives: [
@@ -341,13 +349,7 @@ describe('addHasManyKey', () => {
           isNullable: false,
           directives: [],
         },
-        {
-          name: 'bar',
-          type: 'Bar',
-          isList: true,
-          isNullable: true,
-          directives: [{ name: 'hasMany', arguments: {} }],
-        },
+        hasManyField,
       ],
     };
     const associationFieldPartitionKey: CodeGenField = {
@@ -389,13 +391,20 @@ describe('addHasManyKey', () => {
       associatedWith: associationFieldPartitionKey,
       associatedWithFields: [associationFieldPartitionKey, associationFieldSortKey],
     };
-    const expectedKeyDirective = { name: 'key', arguments: { name: 'gsi-Foo.fooBarsId', fields: ['fooBarsId', 'fooBarsWarehouseId'] } };
+    const expectedKeyDirective = { name: 'key', arguments: { name: 'gsi-Foo.bar', fields: ['fooBarsId', 'fooBarsWarehouseId'] } };
 
-    addHasManyKey(fooModel, connection);
+    addHasManyKey(hasManyField, fooModel, connection);
     expect(barModel.directives).toEqual(expect.arrayContaining([expectedKeyDirective]));
   });
 
   it('works with non-cpk fallback connection for empty list', () => {
+    const hasManyField: CodeGenField = {
+      name: 'bar',
+      type: 'Bar',
+      isList: true,
+      isNullable: true,
+      directives: [{ name: 'hasMany', arguments: {} }],
+    };
     const fooModel: CodeGenModel = {
       name: 'Foo',
       type: 'model',
@@ -410,13 +419,7 @@ describe('addHasManyKey', () => {
           isNullable: false,
           directives: [],
         },
-        {
-          name: 'bar',
-          type: 'Bar',
-          isList: true,
-          isNullable: true,
-          directives: [{ name: 'hasMany', arguments: {} }],
-        },
+        hasManyField,
       ],
     };
     const associationField: CodeGenField = {
@@ -450,13 +453,20 @@ describe('addHasManyKey', () => {
       associatedWith: associationField,
       associatedWithFields: [],
     };
-    const expectedKeyDirective = { name: 'key', arguments: { name: 'gsi-Foo.fooBarsId', fields: ['fooBarsId'] } };
+    const expectedKeyDirective = { name: 'key', arguments: { name: 'gsi-Foo.bar', fields: ['fooBarsId'] } };
 
-    addHasManyKey(fooModel, connection);
+    addHasManyKey(hasManyField, fooModel, connection);
     expect(barModel.directives).toEqual(expect.arrayContaining([expectedKeyDirective]));
   });
 
   it('works with non-cpk fallback connection for null assocatedWithFields', () => {
+    const hasManyField: CodeGenField = {
+      name: 'bar',
+      type: 'Bar',
+      isList: true,
+      isNullable: true,
+      directives: [{ name: 'hasMany', arguments: {} }],
+    };
     const fooModel: CodeGenModel = {
       name: 'Foo',
       type: 'model',
@@ -471,13 +481,7 @@ describe('addHasManyKey', () => {
           isNullable: false,
           directives: [],
         },
-        {
-          name: 'bar',
-          type: 'Bar',
-          isList: true,
-          isNullable: true,
-          directives: [{ name: 'hasMany', arguments: {} }],
-        },
+        hasManyField,
       ],
     };
     const associationField: CodeGenField = {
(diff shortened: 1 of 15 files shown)


packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
```diff
@@ -910,6 +920,11 @@ export class AppSyncModelVisitor<
             } else {
               addFieldToModel(connectionInfo.connectedModel, connectionInfo.associatedWith);
             }
+            // Add the key to the connected model if it's not explicitly defined
+            //   (either via @index or @belongsTo)
+            if (shouldImputeKeyForUniDirectionalHasMany && hasManyHasImplicitKey(model, connectionInfo)) {
```
AaronZyLee, line 925: Since this is within a for loop, we can take advantage of the iterator `field` and pass it into the function. The `field` here should already has `hasMany` attached.

packages/appsync-modelgen-plugin/src/utils/process-has-many.ts
```diff
@@ -134,3 +135,82 @@ export function getConnectedFieldsForHasMany(
...
+
+/**
+ * This is a bit backwards, but we need to check if this connection directive specifies an index.
+ */
+function doesHasManySpecifyIndexName(model: CodeGenModel, hasManyConnection: CodeGenFieldConnectionHasMany): boolean {
+  const directiveIsHasMany: DirectiveFiter = (d) => d.name === TransformerV2DirectiveName.HAS_MANY;
+  const directiveHasIndexName: DirectiveFiter = (d) => d.arguments.indexName;
+  const fieldHasHasManyDirectiveWithIndex: FieldFilter = (f) => f.directives.some(d => directiveIsHasMany(d) && directiveHasIndexName(d));
+  const fieldMatchesConnectedModelName: FieldFilter = (f) => f.type === hasManyConnection.connectedModel.name;
+  return model.fields.some(f => fieldMatchesConnectedModelName(f) && fieldHasHasManyDirectiveWithIndex(f));
```
AaronZyLee, line 192: This will probably return false positive if there are multiple `hasMany` with the same child type. Still I am doubt about its accuracy of this kind of schema. 
```GraqhQL
type Post @model {
  comments1: [Comment] @hasMany
  comments2: [Comment] @hasMany(indexName: "byComment2")
}
type Comment @model {
  content: String
}
```

AaronZyLee: One solution is that instead of finding fields we can directly use the field from the loop as mentioned in the base visitor

packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
alharris-at: Yeah, that's a good point, can simplify the check a bit, and resolve the false positive issue you brought up below. I'll update.

packages/appsync-modelgen-plugin/src/utils/process-has-many.ts
alharris-at: Makes sense, yeah leveraging the field from the visitor should mitigate the need for this matcher.

packages/appsync-modelgen-plugin/src/__tests__/utils/process-has-many.test.ts
```diff
@@ -0,0 +1,382 @@
...
+      ],
+    }
+    const connection: CodeGenFieldConnectionHasMany = {
+      kind: CodeGenConnectionType.HAS_MANY,
+      connectedModel: barModel,
+      isConnectingFieldAutoCreated: true,
+      associatedWith: associationField,
+      associatedWithFields: [associationField],
+    };
+    expect(hasManyHasImplicitKey(hasManyField, fooModel, connection)).toBeTruthy();
```
phani-srikar, line 74: nit: here and in other places, it's better to avoid `toBeTruthy` since it means "type conversion of the value to boolean is true" and hence differs from `toBeTrue` or `toBe(true)`.

packages/appsync-modelgen-plugin/src/__tests__/utils/process-has-many.test.ts
```diff
@@ -0,0 +1,382 @@
...
+    expect(hasManyHasImplicitKey(hasManyField, fooModel, connection)).toBeFalsy();
+  });
+
+  it('returns false for explicit index and id', () => {
+    /**
+     * Example using the following schema:
+     *
+     * type Foo @model {
+     *   id: ID!
+     *   bar: [Bar] @hasMany(indexName: "byFoo", fields: ["id"])
```
phani-srikar, line 162: what happens in case we have only `fields` but no `indexName` argument? It should still add the `key` right?

phani-srikar (changes requested):
couple of nits

packages/appsync-modelgen-plugin/src/utils/process-has-many.ts
```diff
@@ -134,3 +135,72 @@ export function getConnectedFieldsForHasMany(
...
+/**
+ * Extract the name and list of keys from the connection information, and add the key to
+ * the related model.
+ */
+export function addHasManyKey(hasManyConnection: CodeGenFieldConnectionHasMany): void {
+  const associatedFieldNames = getConnectionAssociatedFields(hasManyConnection).map(f => f.name);
+  const connectedModel = hasManyConnection.connectedModel;
+  // Applying consistent auto-naming as the transformer does today
+  // https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-graphql-relational-transformer/src/resolvers.ts#L334-L396
+  const name = `gsi-${connectedModel.name}.${associatedFieldNames[0]}`;
```
AaronZyLee, line 204: I just test locally and find that this naming in transformer should be `gsi-${parent-model.name}.${hasManyFieldName}`
Take the testing schema as example
```GraphQL
type Project @model {
  id: ID! @primaryKey(sortKeyFields: ["name"])
  name: String!
  teams: [Team] @hasMany
}
type Team @model {
  name: String!
}
```
The GSI name is `gsi-Project.teams`. (Confirmed with the `build/stacks/Team.json` in the api folder)

AaronZyLee: The CPK feature flag also changes a behavior for the auto-generated index. When the feature flag is true, it will add  all the associated fields in the index component(aka `projectTeamsId` and `projectTeamsName` in the example). When it is false, it only uses the primary key field(`projectTeamId`) for the index (no sort key fields defined). 
I notice `getConnectionAssociatedFields` uses the same logic. We can add a unit test to verify this scenario.

packages/appsync-modelgen-plugin/src/__tests__/utils/process-has-many.test.ts
alharris-at: I'll update.

packages/appsync-modelgen-plugin/src/__tests__/utils/process-has-many.test.ts
alharris-at: It's not a valid use-case. We throw an InvalidDirectiveError in that case, since it's registered as a Partial sort key (which isn't exactly right either, but it's correctly erroring).

packages/appsync-modelgen-plugin/src/utils/process-has-many.ts
alharris-at: I actually do have a case that flexes that behavior, in the `process-has-many` tests, the `works with non-cpk fallback connection` case should cover this, by providing an empty associatedWithFields list. I'll verify I can pass a null in as well.

alharris-at:
@phani-srikar / @AaronZyLee responded to comments, and made some updates accordingly. Please take another look ❤️

packages/appsync-modelgen-plugin/src/utils/process-has-many.ts
```diff
@@ -134,3 +135,72 @@ export function getConnectedFieldsForHasMany(
...
+  const hasCorrespondingBelongsTo = doesHasManyConnectionHaveCorrespondingBelongsTo(model, hasManyConnection);
+  const hasIndexNameSpecified = doesHasManySpecifyIndexName(field);
+  return !(hasCorrespondingBelongsTo || hasIndexNameSpecified);
+}
+
+/**
+ * Extract the name and list of keys from the connection information, and add the key to
+ * the related model.
+ */
+export function addHasManyKey(model: CodeGenModel, hasManyConnection: CodeGenFieldConnectionHasMany): void {
```
AaronZyLee, line 199: The field here should be changed as well to match the naming in transformer.
```TypeScript
export function addHasManyKey(field: CodeGenField, model: CodeGenModel, hasManyConnection: CodeGenFieldConnectionHasMany): void {
  const associatedFieldNames = getConnectionAssociatedFields(hasManyConnection).map(f => f.name);
  const connectedModel = hasManyConnection.connectedModel;
  // Applying consistent auto-naming as the transformer does today
  // https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-graphql-relational-transformer/src/resolvers.ts#L334-L396
  const name = `gsi-${model.name}.${field.name}`; // ----------> instead of associatedFieldNames[0]
  addKeyToModel(connectedModel, name, associatedFieldNames);
}
```
And the statement in base visitor is
```TypeScript
    if (shouldImputeKeyForUniDirectionalHasMany && hasManyHasImplicitKey(field, model, connectionInfo)) {
      addHasManyKey(field, model, connectionInfo);
    }

```

AaronZyLee (changes requested):
Overall LGTM. Just need one more change

packages/appsync-modelgen-plugin/src/utils/process-has-many.ts
alharris-at: Ahh, I missed the field change in your previous comment. I'll update, one moment.

phani-srikar (approved):

AaronZyLee (approved):

PR #513, opened 2022-11-11
ci: add build app e2e test for TypeScript

dpilch:
#### Description of changes

A recent change (https://github.com/aws-amplify/amplify-codegen/pull/489) caused invalid model generation in some cases. This PR introduces E2E tests that attempt to build an app with the generated models. This is starting with a basic set of test using the [schema test library](https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-graphql-schema-test-library/src/index.ts) so that coverage for the error is introduced sooner rather than later.

Future improvements:
* Add build time check for all other platforms
* Add runtime check for all platforms
* Add various feature flags on/off

#### Issue #, if available

PR to introduce the error: https://github.com/aws-amplify/amplify-codegen/pull/489

PR to fix the error: https://github.com/aws-amplify/amplify-codegen/pull/500

Ran tests in circleci: https://app.circleci.com/pipelines/github/aws-amplify/amplify-codegen/2220/workflows/02eb292e-5b68-45f3-a56d-0cd9b2424c98

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-codegen-e2e-core/src/init/cra.ts
```diff
@@ -0,0 +1,36 @@
...
+    const s = { ...defaultSettings, ...settings };
+    const chain = spawn('npm', ['install'], { cwd, stripColors: true, disableCIDetection: s.disableCIDetection });
+
+    chain.run((err: Error) => {
+      if (err) {
+        reject(err);
+      } else {
+        resolve();
+      }
+    });
```
alharris-at, line 19: `runAsync` is a nice way to reduce the if/else here. This block could be rewritten.

```typescript
return spawn('npm', ['install'], { cwd, stripColors: true, disableCIDetection: s.disableCIDetection }).runAsync();
```

alharris-at: And same for the following fn as well.

packages/amplify-codegen-e2e-tests/test-apps/ts/scripts/set-schema.js
```diff
@@ -0,0 +1,19 @@
+const fs = require('fs');
+const { schemas } = require('@aws-amplify/graphql-schema-test-library');
+
+const schemaName = process.argv.slice(2, 3)[0];
+if (!schemaName) {
+  throw new Error('A schema name must be supplied.');
+}
+const schema = schemas[schemaName];
+if (!schema) {
+  throw new Error(`${schemaName} does not exist in the schema test library.`);
```
alharris-at, line 10: Maybe list possible options as well?

alharris-at (commented):
Couple of comments, none of which are blocking.

The only question was whether moving the test app itself out of the `packages` directory resolves the need to have all the lint rules updated, setting it private so we don't accidentally publish, etc.

dpilch:
Lint ignore would still be necessary because the lint script is [targeting all JS/TS files](https://github.com/aws-amplify/amplify-codegen/blob/main/package.json#L9) (we could update this to be more precise though). I think the test app is safe where it is because `amplify-codegen-e2e-tests` itself is also private. I don't really feel strongly on where it is though. I put it here because I though `../../` would be confusing/ugly in the test.

packages/amplify-codegen-e2e-core/src/init/cra.ts
dpilch: Doesn't look like we have a `runAsync` https://github.com/aws-amplify/amplify-codegen/blob/main/packages/amplify-codegen-e2e-core/src/utils/nexpect.ts#L56-L72.

.circleci/config.yml
```diff
@@ -379,6 +379,14 @@ jobs:
     environment:
       TEST_SUITE: src/__tests__/graphql-codegen-ios.test.ts
       CLI_REGION: us-west-1
+  build-app-ts-e2e-test:
```
phani-srikar, line 382: These changes to config should go into `.circleci/config.base.yml` and I believe this file gets overridden by config generated by running `yarn split-e2e-tests`.

packages/amplify-codegen-e2e-tests/test-apps/ts/public/manifest.json
```diff
@@ -0,0 +1,25 @@
...
+  "short_name": "React App",
+  "name": "Create React App Sample",
+  "icons": [
+    {
+      "src": "favicon.ico",
+      "sizes": "64x64 32x32 24x24 16x16",
+      "type": "image/x-icon"
+    },
+    {
+      "src": "logo192.png",
```
phani-srikar, line 11: do you really need these images in your app?

packages/amplify-codegen-e2e-tests/test-apps/ts/src/setupTests.ts
```diff
@@ -0,0 +1,5 @@
+// jest-dom adds custom jest matchers for asserting on DOM nodes.
+// allows you to do things like:
+// expect(element).toHaveTextContent(/react/i)
+// learn more: https://github.com/testing-library/jest-dom
+import '@testing-library/jest-dom';
```
phani-srikar, line 5: blank file?

phani-srikar (changes requested):
Looks good, can you take another pass and remove any boiler plate files from react app? Thanks.

packages/amplify-codegen-e2e-tests/test-apps/ts/public/manifest.json
dpilch: Removed

packages/amplify-codegen-e2e-tests/test-apps/ts/src/setupTests.ts
dpilch: Removed

packages/amplify-codegen-e2e-tests/test-apps/ts/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "private": true,
+  "dependencies": {
+    "@testing-library/jest-dom": "^5.14.1",
+    "@testing-library/react": "^13.0.0",
+    "@testing-library/user-event": "^13.2.1",
+    "@types/jest": "^27.0.1",
+    "@types/node": "^16.7.13",
+    "@types/react": "^18.0.0",
+    "@types/react-dom": "^18.0.0",
+    "aws-amplify": "^4.3.43",
```
AaronZyLee, line 13: TODO reminder: we need to bump the version if there is a major version bump for `aws-amplify`

packages/amplify-codegen-e2e-tests/test-apps/ts/.graphqlconfig.yml
```diff
@@ -0,0 +1,15 @@
+projects:
```
AaronZyLee, line 1: This file can be removed if doc/type gen is not in scope of this test.

packages/amplify-codegen-e2e-tests/test-apps/ts/scripts/set-schema.js
```diff
@@ -0,0 +1,24 @@
+const fs = require('fs');
```
AaronZyLee, line 1: Is this script in use? I see there is a method `updateApiSchemaWithText` in e2e test used for setting up the schema.

AaronZyLee (changes requested):

packages/amplify-codegen-e2e-tests/test-apps/ts/scripts/set-schema.js
dpilch: This is used for local testing outside of the test suite. It is not strictly necessary, but I found it useful.

packages/amplify-codegen-e2e-tests/test-apps/ts/package.json
dpilch: This should already be v5 actually. Good catch

dpilch: Changed to `latest` so that it is always up to date.

AaronZyLee (approved):

alharris-at (approved):

PR #520, opened 2022-11-29
ci: add build e2e test for swift

dpilch:
#### Description of changes

See https://github.com/aws-amplify/amplify-codegen/pull/513 for similar test for TypeScript apps.

#### Issue #, if available

PR to introduce the error: https://github.com/aws-amplify/amplify-codegen/pull/489

PR to fix the error: https://github.com/aws-amplify/amplify-codegen/pull/500

#### Description of how you validated changes

[Run on CircleCI](https://app.circleci.com/pipelines/github/aws-amplify/amplify-codegen/2643/workflows/92d3cc95-6956-4f8a-ae20-9235e61d940b)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed or added](https://github.com/aws-amplify/amplify-codegen/blob/main/CONTRIBUTING.md#tests)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #524, opened 2022-12-02
chore: upgrade to graphql 15

dpilch:
#### Description of changes

Upgrade `graphql` package to be inline with `graphql` version used in the CLI. 

https://github.com/aws-amplify/amplify-cli/blob/dev/packages/amplify-cli/package.json#L91

In some rare cases both graphql 14 and 15 were resolved in the package leading to errors.

```
Error: Cannot use GraphQLScalarType \\"String\\" from another module or realm.

Ensure that there is only one instance of \\"graphql\\" in the node_modules
directory. If different versions of \\"graphql\\" are the dependencies of other
relied on modules, use \\"resolutions\\" to ensure only one version is installed.

https://yarnpkg.com/en/docs/selective-version-resolutions

Duplicate \\"graphql\\" modules cannot be used at the same time since different
versions may have different capabilities and behavior. The data from one
version used in the function from another could produce confusing and
spurious results.
```

#### Issue #, if available

The above error would surface when the CLI was installed via Verdaccio on the macos executor on CircleCI.

https://github.com/aws-amplify/amplify-codegen/pull/520

* Unit tests
* E2E tests
* **Further testing is still needed**

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


packages/appsync-modelgen-plugin/package.json
```diff
@@ -40,12 +40,13 @@
   },
   "devDependencies": {
     "@graphql-codegen/testing": "^1.17.7",
-    "graphql": "^14.5.8",
+    "@graphql-codegen/typescript": "^2.8.3",
```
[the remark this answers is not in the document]
dpilch: Test would fail with the following without including this.
```
FAIL src/__tests__/visitors/appsync-javascript-visitor.test.ts
  ● Test suite failed to run

    Cannot find module '@graphql-codegen/typescript' from '../../node_modules/@graphql-codegen/testing/cjs/resolvers-common.js'
```

I think GraphQL 14 may have included this dependency before.

dpilch:
Planning for an official release. There is already a tagged release ([3.3.5-graphql15.0](https://www.npmjs.com/package/amplify-codegen/v/3.3.5-graphql15.0)) used for testing.

phani-srikar (approved):

packages/graphql-types-generator/test/__snapshots__/jsonOutput.ts.snap
```diff
@@ -49,7 +49,8 @@ exports[`JSON output should generate JSON output for a mutation with an enum and
 							\\"responseName\\": \\"__typename\\",
 							\\"fieldName\\": \\"__typename\\",
 							\\"type\\": \\"String!\\",
-							\\"isConditional\\": false
+							\\"isConditional\\": false,
+							\\"isDeprecated\\": false
```
phani-srikar, line 53: why is this extra field being added in the tests?

phani-srikar (changes requested):

packages/graphql-types-generator/test/__snapshots__/jsonOutput.ts.snap
dpilch: `isDepreacted` became required on `GraphQLField` in version 15. So it now shows up on all fields including `__typename`. https://github.com/graphql/graphql-js/pull/2469

I don't see this causing any issue though.

phani-srikar (approved):

PR #530, opened 2023-01-06
fix(appsync-modelgen-plugin): associatedWith fields in HasOne bi-directional with CPK enabled

phani-srikar:
#### Description of changes

When the CPK feature flag is `ON`, this change fixes the `associatedWith` fields that are generated in a Bi-Directional Has-One relationship. 
We add an option `shouldUseFieldsInAssociatedWithInHasOne` to control this change and limit it only to the Model Introspection Visitor (produces `model-introspection.json` output), JS, TS and JSON metadata vistors (linked to `schema.js` output).

Unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/appsync-modelgen-plugin/src/utils/process-connections-v2.ts
+++ b/packages/appsync-modelgen-plugin/src/utils/process-connections-v2.ts
@@ -129,14 +129,15 @@ export function processConnectionsV2(
   model: CodeGenModel,
   modelMap: CodeGenModelMap,
   shouldUseModelNameFieldInHasManyAndBelongsTo: boolean = false,
-  isCustomPKEnabled: boolean = false
+  isCustomPKEnabled: boolean = false,
+  shouldUseFieldsInAssociatedWithInHasOne: boolean = false,
 ): CodeGenFieldConnection | undefined {
   const connectionDirective = field.directives.find(d => d.name === 'hasOne' || d.name === 'hasMany' || d.name === 'belongsTo');
 
   if (connectionDirective) {
     switch (connectionDirective.name) {
       case 'hasOne':
-        return processHasOneConnection(field, model, modelMap, connectionDirective, isCustomPKEnabled);
+        return processHasOneConnection(field, model, modelMap, connectionDirective, isCustomPKEnabled, shouldUseFieldsInAssociatedWithInHasOne);
       case 'belongsTo':
         return processBelongsToConnection(field, model, modelMap, connectionDirective, isCustomPKEnabled);
       case 'hasMany':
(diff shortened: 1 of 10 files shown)


packages/appsync-modelgen-plugin/src/utils/process-connections-v2.ts
```diff
@@ -129,14 +129,15 @@ export function processConnectionsV2(
   model: CodeGenModel,
   modelMap: CodeGenModelMap,
   shouldUseModelNameFieldInHasManyAndBelongsTo: boolean = false,
-  isCustomPKEnabled: boolean = false
+  isCustomPKEnabled: boolean = false,
+  shouldUseFieldsInAssociatedWithInHasOne: boolean = false,
```
alharris-at, line 133: Naming n.p. but you probably want `Within` instead of `WithIn` here.

packages/appsync-modelgen-plugin/src/utils/process-has-one.ts
```diff
@@ -59,6 +60,6 @@ export function processHasOneConnection(
  * @param otherSide other side model of hasOne connection
```
alharris-at, line 60: Could you document the new param here?

alharris-at (dismissed):
Generally LGTM - a few nits on naming, but none blocking approval.

packages/appsync-modelgen-plugin/src/utils/process-connections-v2.ts
phani-srikar: This should be read as "should use fields in associatedWith in hasOne".

alharris-at (dismissed):

AaronZyLee (dismissed):
LGTM :shipit:

david-mcafee (dismissed):
LGTM!

packages/appsync-modelgen-plugin/src/utils/process-connections-v2.ts
```diff
@@ -129,14 +129,15 @@ export function processConnectionsV2(
...
 ): CodeGenFieldConnection | undefined {
   const connectionDirective = field.directives.find(d => d.name === 'hasOne' || d.name === 'hasMany' || d.name === 'belongsTo');
 
   if (connectionDirective) {
     switch (connectionDirective.name) {
       case 'hasOne':
-        return processHasOneConnection(field, model, modelMap, connectionDirective, isCustomPKEnabled);
+        return processHasOneConnection(field, model, modelMap, connectionDirective, isCustomPKEnabled, shouldUseFieldsInAssociatedWithInHasOne);
```
sundersc, line 140: How do we compute the value for `shouldUseFieldsInAssociatedWithInHasOne`?

phani-srikar: This value is set in the top level visitors and if not it defaults to `false`. In this PR, we only set this to `true` for Model Introspection visitor which generates the introspection schema.

sundersc (dismissed):

alharris-at (approved):

AaronZyLee (approved):

PR #542, opened 2023-02-03
fix(modelgen-swift): respect index sk fields in associate fields

### LINKED ISSUE #539: Swift Model Codegen for Unidirectional Has-Many AssociatedFields Missing Index
This doesn't look accurate

_Originally posted by @lawmicha in https://github.com/aws-amplify/amplify-codegen/pull/538#discussion_r1095218053_

AaronZyLee:
#### Description of changes

- The `associatedWith` fields now respect the index sortKeyFields in explicit uni hasMany relation. This change only affects swift modelgen.
- Rewrite the flag value parsing for `processDirectives`. Add new type definition and default value for directive process configuration.

#### Issue #, if available

Fix #539 

`yarn test`
`amplify-dev codegen models`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/appsync-modelgen-plugin/src/utils/process-has-many.ts
+++ b/packages/appsync-modelgen-plugin/src/utils/process-has-many.ts
@@ -107,7 +107,8 @@ export function getConnectedFieldsForHasMany(
     if (!otherSideConnectedField) {
       throw new Error(`Can not find key field ${otherSideConnectedFieldName} in ${connectedModel.name}`);
     }
-    return [otherSideConnectedField];
+    const sortKeyFieldNames: string[] = otherSideConnectedDir?.arguments.sortKeyFields ?? [];
+    return [otherSideConnectedField, ...sortKeyFieldNames.map(sk => connectedModel.fields.find(f => f.name === sk)!)];
   }
 
   // When fields argument is not defined, auto generate connected fields
--- a/packages/appsync-modelgen-plugin/src/utils/process-has-many.ts
+++ b/packages/appsync-modelgen-plugin/src/utils/process-has-many.ts
@@ -1,4 +1,4 @@
-import { CodeGenDirective, CodeGenField, CodeGenFieldDirective, CodeGenModel, CodeGenModelMap } from '../visitors/appsync-visitor';
+import { CodeGenDirective, CodeGenDirectiveProcessConfig, CodeGenField, CodeGenFieldDirective, CodeGenModel, CodeGenModelMap } from '../visitors/appsync-visitor';
 import { TransformerV2DirectiveName, DEFAULT_HASH_KEY_FIELD } from './constants';
 import { getDirective, getOtherSideBelongsToField } from './fieldUtils';
 import { getModelPrimaryKeyComponentFields } from './fieldUtils';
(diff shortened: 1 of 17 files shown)


packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
```diff
@@ -221,6 +221,28 @@ type ManyToManyContext = {
   directive: CodeGenDirective;
 };
 
+// Configuration used for processing directives
+export type CodeGenDirectiveProcessConfig = {
+  // This flag is going to be used for using custom primary key feature
+  isCustomPKEnabled: boolean,
```
phani-srikar, line 227: does this mean that there is a `@primaryKey` present in the schema or just that if it's present, we will codegen differently?

packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
```diff
@@ -221,6 +221,28 @@ type ManyToManyContext = {
...
+  // This flag is going to be used to tight-trigger on JS implementations only.
+  shouldImputeKeyForUniDirectionalHasMany: boolean,
+  // This flag is currently used in JS/TS and Model introspection generation only.
+  shouldUseFieldsInAssociatedWithInHasOne: boolean,
+  // This flag is currently used in Swift only to have sort key fields of index included in associatedWith fields
+  shouldRespectSortKeyFieldsOfIndexInAssociatedWithFields:boolean,
+}
+
+export const defaultCodegenDirectiveProcessConfig: CodeGenDirectiveProcessConfig = {
+  isCustomPKEnabled: true,
```
phani-srikar, line 239: The ideal default here would be `this.isCustomPKEnabled()` but since this is defined outside the class, can you specify why we chose `true` as default for this one and not `false` like others below?

packages/appsync-modelgen-plugin/src/utils/process-has-many.ts
```diff
@@ -107,7 +108,10 @@ export function getConnectedFieldsForHasMany(
     if (!otherSideConnectedField) {
       throw new Error(`Can not find key field ${otherSideConnectedFieldName} in ${connectedModel.name}`);
     }
-    return [otherSideConnectedField];
+    const sortKeyFieldNames: string[] = otherSideConnectedDir?.arguments.sortKeyFields ?? [];
+    return shouldRespectSortKeyFieldsOfIndexInAssociatedWithFields
+      ? [otherSideConnectedField, ...sortKeyFieldNames.map(sk => connectedModel.fields.find(f => f.name === sk)!)]
```
phani-srikar, line 113: Is the `!` necessary in the `find`?

phani-srikar (changes requested):
Minor comemnts

packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
AaronZyLee: This is determined by the feature flag value rather than the presence of the directive. If the value is true and the schema is a CPK (the @primaryKey is attached to non-id field/id PK with SK defined), it will codegen differently

packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
AaronZyLee: I just change the default value for CPK to false. This will be override by each visitor's own method of `this.isCustomPKEnabled()`

phani-srikar (approved):

PR #550, opened 2023-02-28
fix: E2e testing

AaronZyLee:
#### Description of changes

- Changes the structure of build-test-deploy workflow
  - Removed the build and publish cli jobs on macOS due to the persist to workspace for concurrent jobs, which results in the downstream `deploy` job to fail at `attach_workspace` step. See [details](https://app.circleci.com/pipelines/github/aws-amplify/amplify-codegen/2763/workflows/80d3fd2d-6c31-479d-9f28-490a27182131/jobs/14569)
  - Added additional steps in `build-swift-app` test for environment setup
  - Example worlflow
    - Before: https://app.circleci.com/pipelines/github/aws-amplify/amplify-codegen/2756/workflows/11bb6e1f-6e30-452f-9548-21b4bd72b1da
    - After: https://app.circleci.com/pipelines/github/aws-amplify/amplify-codegen/2783/workflows/39463853-56b9-4f32-aa27-171f616928d1
- Added new region and walkthrough in e2e core for `amplify configure`
- Use the `release` branch instead of main for `build-swift-app` test

Note: the `build-swift-app` is still failing on the CLI of v10.8.1. This is expected as the xcode script of importing models is not resolved correctly due to the package renaming. The fixing PR is out in CLI and will go into next release.
https://github.com/aws-amplify/amplify-cli/pull/12145

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-codegen-e2e-core/src/configure/index.ts
+++ b/packages/amplify-codegen-e2e-core/src/configure/index.ts
@@ -55,8 +55,6 @@ export function amplifyConfigure(settings: AmplifyConfiguration): Promise<void> 
     singleSelect(chain, s.region, amplifyRegions);
 
     chain
-      .wait('user name:')
-      .sendCarriageReturn()
       .wait('Press Enter to continue')
       .sendCarriageReturn()
       .wait('accessKeyId')
--- a/packages/amplify-codegen-e2e-core/src/configure/index.ts
+++ b/packages/amplify-codegen-e2e-core/src/configure/index.ts
@@ -19,6 +19,7 @@ export const amplifyRegions = [
   'us-west-1',
   'us-west-2',
   'eu-north-1',
+  "eu-south-1",
   'eu-west-1',
   'eu-west-2',
   'eu-west-3',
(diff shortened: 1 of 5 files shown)


packages/amplify-codegen-e2e-core/src/configure/index.ts
```diff
@@ -19,6 +19,7 @@ export const amplifyRegions = [
   'us-west-1',
   'us-west-2',
   'eu-north-1',
+  "eu-south-1",
```
sundersc, line 22: In Data repo, I noticed this change need to sync exactly with what's in cloudformation provider CLI package. Otherwise the select E2E helper selects a wrong region. Please make sure that this isn't the case with codegen.

sundersc (commented):
Overall LGTM. Just want to make sure that adding the region won't end up picking a wrong region in the dropdown.

packages/amplify-codegen-e2e-core/src/configure/index.ts
AaronZyLee: The new region is added in v10.8.0 and that is the point the e2e tests start to fail. As the codegen uses the latest cli-internal, this fix will be synced with the cfn provider

AaronZyLee: Before this fix, the wrong region was picked when using CLI version based on 10.8.0

alharris-at (approved):
Change seems reasonable, discussed offline about verifying that Swift test passes before pushing.

PR #557, opened 2023-03-09
fix(ci): rm suffix of os in cache

AaronZyLee:
#### Description of changes

Removed the OS suffix for cache naming as there is a mismatch between `parameter.os` (which is a wrapped object rather than single string and its alias `l`/`w`. This causes the cache not restoring issue and the deploy job to fail.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/.circleci/config.base.yml
+++ b/.circleci/config.base.yml
@@ -82,11 +82,11 @@ jobs:
       - run: yarn cache clean --force
       - run: yarn run production-build
       - save_cache:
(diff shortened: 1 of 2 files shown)


alharris-at (approved):

phani-srikar (approved):
I've seen some issue before when we restore the cache stored from Linux on Windows and vice-versa. But that was way back. If this is working end to end, LGTM.

PR #570, opened 2023-04-12
fix(introspection): primary key info when pk is belongsTo field

AaronZyLee:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-model-introspection-visitor.test.ts.snap
+++ b/packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-model-introspection-visitor.test.ts.snap
@@ -1,5 +1,1529 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Custom primary key tests should generate correct model intropection file validated by JSON schema and not throw error when custom PK is disabled 1`] = `
+"{
+    \\"version\\": 1,
+    \\"models\\": {
+        \\"Project1\\": {
+            \\"name\\": \\"Project1\\",
+            \\"fields\\": {
+                \\"projectId\\": {
+                    \\"name\\": \\"projectId\\",
+                    \\"isArray\\": false,
+                    \\"type\\": \\"ID\\",
+                    \\"isRequired\\": true,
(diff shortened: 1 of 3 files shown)


packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-model-introspection-visitor.test.ts.snap
```diff
@@ -1,5 +1,1529 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Custom primary key tests should generate correct model intropection file validated by JSON schema and not throw error when custom PK is disabled 1`] = `
```
phani-srikar, line 3: I don't think such a big snapshot makes sense to be added and is feasible to review. Please come up with an alternative test.

packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
```diff
@@ -1028,7 +1028,9 @@ export class AppSyncModelVisitor<
             connectionInfo &&
             connectionInfo.kind !== CodeGenConnectionType.HAS_MANY &&
             connectionInfo.kind !== CodeGenConnectionType.HAS_ONE &&
-            connectionInfo.targetName !== 'id'
+            connectionInfo.targetName !== 'id' &&
+            !(this.config.target === 'introspection' &&
```
phani-srikar, line 1032: What are the other targets that are valid here and why this fix only required for "introspection" target?

phani-srikar (changes requested):
Please include some context in the PR description as it is not clear from the title what this is accomplishing.

packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts
AaronZyLee: The other targets are controlled by the feature flag value. For introspection, there are cases where the app has CPK value as false and the code block above is entered. That's why it is protected not to execute the line when target is introspection and belongsTo connnection field is primary key

phani-srikar (approved):

Issue #445, opened 2022-06-28
API types contain __typename, queries do not


eettaa:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-codegen/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-codegen/blob/master/CONTRIBUTING.md#bugs).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

npm 

### If applicable, what version of Node.js are you using?

14.18.1

### Amplify CLI Version

7.6.21 (sorry, I'm blocked by https://github.com/aws-amplify/amplify-category-api/issues/560)

### What operating system are you using?

Ubuntu

### Amplify Codegen Command

codegen

### Describe the bug

API.ts as generated by codegen has typescript types that include a string literal __typename field, e.g.
```
export type GetSavingsAccountQuery = {
  getSavingsAccount?:  {
    __typename: "SavingsAccount",
    accountId: string
    ...
}
```
But the associated graphql query in queries.ts does NOT query the __typename field:
```
  query GetSavingsAccount($accountId: ID!) {
    getSavingsAccount(accountId: $accountId) {
      accountId
      (no __typename here!)
      ...
  }
}
```

This means that trying to access __typename on a query response will always typecheck but never succeed at runtime (because that field is never returned).

Some frontend graphql clients like ApolloClient automatically append __typename to all queries, which works around this problem.  however, if you were to use the literal query as generated, it would not match the type as generated.

Is there an option to have codegen add __typename to all the queries?

### Expected behavior

all queries in queries.ts also query the __typename field, or (less preferrable to me), API.ts does not export __typename as a field in all query responses.

### Reproduction steps

Take any trivial @model-annotated type in schema
Codegen it
Inspect API.ts and queries.ts, notice one has __typename and one does not.

### GraphQL schema(s)

I have instead included a graphqlconfig.yml:
```
projects:
  projectname:
    schemaPath: path/to/schmea
    includes:
      - path/to/codegen/**/*.ts
      - path/to/custom/**/*.ts
    excludes:
      - ./amplify/**
    extensions:
      amplify:
        codeGenTarget: typescript
        generatedFileName: path/to/codegen/API.ts
[...]

PR #575, opened 2023-04-19
feat: add __typename to selection set

dpilch:
#### Description of changes

**BREAKING CHANGE**

Add `__typename` to selection set of generated queries for object types.

[Typename Introspection](https://spec.graphql.org/June2018/#sec-Type-Name-Introspection) is included in the GraphQL spec and is accessible on the meta field `__typename`. `__typename` is available on object, interface, and union types.

This change adds `__typename` to only object types because we do not generate resolvers for interface and union types. (We could add this anyway for customers not using the transformers).

AppSync is guaranteed to supply `__typename` for all queries not using interface or union types regardless of the resolver implementation. Therefore, this change is only potential breaking for customers meeting the following criteria:
1. using a non-AppSync backend
2. the backend is non-compliant with the GraphQL spec.

Customers will have the option to disable generation of `__typename` with `typenameIntrospection` field in `.graphqlconfig.yml`. The option will default to true when omitted.

```yml
projects:
  typename:
    extensions:
      amplify:
        typenameIntrospection: true
```

#### Issue #, if available

Resolves https://github.com/aws-amplify/amplify-codegen/issues/445

* Unit testing
* `amplify-dev codegen statements` in typescript, ios, and android app and try queries.
    * `__typename` is not accessible on ios and android because the generated classes do no have a `__typename` property, but the inclusion of `__typename` in the query does not cause errors.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/graphql-docs-generator/src/generator/getFields.ts
+++ b/packages/graphql-docs-generator/src/generator/getFields.ts
@@ -22,7 +22,7 @@ export default function getFields(
   field: GraphQLField<any, any>,
   schema: GraphQLSchema,
   depth: number = 2,
-  options: GQLDocsGenOptions
+  options: GQLDocsGenOptions,
 ): GQLTemplateField {
   const fieldType: GQLConcreteType = getType(field.type);
   const renderS3FieldFragment = options.useExternalFragmentForS3Object && isS3Object(fieldType);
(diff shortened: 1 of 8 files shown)


packages/graphql-docs-generator/src/generator/getFields.ts
```diff
@@ -40,6 +40,17 @@ export default function getFields(
       return getFields(subField, schema, adjustDepth(subField, depth), options);
     })
     .filter(f => f);
+
+  // add __typename to selection set.
+  // getFields() does not include __typename because __typename is implicitly included on all object types.
```
phani-srikar, line 45: This assumption might not be valid when we accept any valid GraphQL introspection schema (not generated using our transformers) from the customer that could have this field. Would be more inclined to have a check if this exists and then add it.

packages/graphql-docs-generator/src/generator/getFields.ts
```diff
@@ -40,6 +40,17 @@ export default function getFields(
...
+  if (isObjectType(fieldType)) {
```
phani-srikar, line 46: Do Interfaces also have this meta field? If yes we might need to modify this to `isObjectType(fieldType) || isInterfaceType(fieldType)`.

phani-srikar (changes requested):

phani-srikar (approved):

PR #562, opened 2023-03-27
Refactor GraphQL documents generator package to be browser compatible and Amplify CLI agnostic

phani-srikar:
#### Description of changes

Following changes are made to the GraphQL documents generator package:
1. Moves the steps to read the input GraphQL schema from user's file system to `amplify-codegen` package.
2. Also moves the step to write back the generated GraphQL operations to user's file system to `amplify-codegen` package.
3. GraphQL documents generator package now accepts the GraphQL SDL/introspection schema as a string and returns a map of generated operations that looks like:
```
{
    queries: {"getTodo": "query GetTodo ....", "listTodos": "query ListTodos..."},
    mutations: {"createTodo": "mutation CreateTodo ...."},
    subscriptions: {"onCreateTodo": "subscription onCreateTodo ...."},
    fragments: {"myfragment": "fragment myFragment ...."}
}
```
4. The GraphQL documents generator package can now be used outside of Amplify CLI project and also in a browser environment. Updated README file to reflect the same.

In Subsequent PR:
- Update the logic in `amplify-codegen` plugin to consume this updated API.

#### Description of how you validated changes
Unit testing, 
E2e Testing - Added a React App based browser test. Refer to the checks on the PR for latest status.
running `amplify-dev codegen statements` in iOS sample app

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed or added](https://github.com/aws-amplify/amplify-codegen/blob/main/CONTRIBUTING.md#tests)
- [ ] Relevant documentation is changed or added (and PR referenced)
- [ ] Breaking changes to existing customers are released behind a feature flag or major version update
- [ ] Changes are tested using sample applications for all relevant platforms (iOS/android/flutter/Javascript) that use the feature added/modified

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #583, opened 2023-05-01
feat: consume updated docgen API

phani-srikar:
#### Description of changes

Following up on earlier PR that modifies the documents generator package API, 
this PR ensures the top level codegen plugin is updated with these changes. Here is a summary of changes:
- Handle frontend language specific formatting including "prettifying" the output.
- Handle writing back the generated output to disk.
- Handle reading the input introspection schema from CLI specific location.

Update unit tests.
Sample apps for JS(JS/TS/Flow), iOS(GraphQL) to verify for a schema with `@hasOne-@belongsTo` relationship that output using public version of Codegen and `dev` version produce the same output. Only difference b/w outputs is for `GraphQL` targets where a new line is added b/w generated operations for better readability.
[Link to full e2e](https://app.circleci.com/pipelines/github/aws-amplify/amplify-codegen/3035/workflows/1bce340a-1592-4a9a-b32f-1d26d7303a4d)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/graphql-docs-generator/src/generator/utils/getSchemaType.ts
```diff
@@ -5,7 +5,6 @@ import { SchemaType } from '../types';
   which are the current supported formats.
```
phani-srikar, line 5: Addresses comments on previous PR https://github.com/aws-amplify/amplify-codegen/pull/562#pullrequestreview-1402957715

packages/amplify-codegen/src/commands/statements.js
```diff
@@ -50,45 +50,52 @@ async function generateStatements(context, forceDownloadSchema, maxDepth, withou
...
+    const ops = generatedStatements[op];
+    if (ops && ops.size) {
+      const formattedStatements = (new GraphQLStatementsFormatter(language)).format(ops);
+      const outputFile = path.resolve(path.join(outputPath, `${op}.${fileExtension}`));
+      fs.writeFileSync(outputFile, formattedStatements);
+    }
+  });
+
+  if (fileExtension === 'graphql') {
+    // External Fragments are rendered only for GraphQL targets
```
AaronZyLee, line 93: Is this comment originated from the doc-gen implementation? Do we prevent fragments from writing to file system when they are  generated under non-GQL language? If not, I suggest the fragments part to be merged to the code block above as it will check the `ops && ops.size` anyway and use the same format class/methods

packages/amplify-codegen/src/utils/GraphQLStatementsFormatter.js
```diff
@@ -0,0 +1,72 @@
+const prettier = require('prettier');
```
AaronZyLee, line 1: I suggest a unit test (snapshot) for this class with a predefined statements for the lint overrides/header comments under each language.

AaronZyLee (changes requested):

packages/amplify-codegen/src/commands/statements.js
phani-srikar: There was no straight forward comment, but if you look at the template for non-graphql language, say, JS for example, you can see that it [uses `renderToVariable` partial](https://github.com/aws-amplify/amplify-codegen/blob/main/packages/graphql-docs-generator/templates/javascript.hbs) which in turn looks like this:
```
{{#each operations }}
  export const {{formatName name}} =  /* GraphQL */`{{#format }}
      {{> renderOp }}
    {{/format}}`;
{{/each}}
```
But for GraphQL, the template looks like:
```
# this is an auto generated file. This will be overwritten
{{#each operations }}
  {{> renderOp }}
{{/each}}
{{#each fragments }}
  {{> renderExternalFragment }}
{{/each}}
```
So for non-graphql languages, we're ignoring the fragments arg that gets passed to the [render method](https://github.com/aws-amplify/amplify-codegen/blob/066615e97c54e9f62b599991e058fc0d70aa22b4/packages/graphql-docs-generator/src/index.ts#L62)

packages/amplify-codegen/src/utils/GraphQLStatementsFormatter.js
```diff
@@ -0,0 +1,72 @@
...
+
+const CODEGEN_WARNING = 'this is an auto generated file. This will be overwritten';
```
dpilch, line 3: nit
```suggestion
const CODEGEN_WARNING = 'This is an auto generated file. This will be overwritten.';
```

packages/graphql-docs-generator/src/generator/utils/getSchemaType.ts
```diff
@@ -14,8 +13,6 @@ export const getSchemaType = (schema: string) => {
     }
   }
   catch {
-    return schemaType;
+    return SchemaType.SDL;
   }
-
-  return schemaType;
```
dpilch, line 20: There will be no return for when the schema is successfully parsed as JSON but does not have a `data` or `__schema` property. What is correct behavior here? Throw an error or default to SDL?

dpilch: Oh I see now that it throws an error later in `buildSchema`. Maybe add to the docstring that returning `undefined` is the intended behavior (but not a blocker).

AaronZyLee (approved):

PR #589, opened 2023-05-09
fix: codegen downgrade to version 3

AaronZyLee:
#### Description of changes

- Revert the `__typename` change in doc generator
- Downgrade amplify codegen to version 3 to avoid version mismatch error from JS datastore

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed or added](https://github.com/aws-amplify/amplify-codegen/blob/main/CONTRIBUTING.md#tests)
- [ ] Relevant documentation is changed or added (and PR referenced)
- [ ] Breaking changes to existing customers are released behind a feature flag or major version update
- [ ] Changes are tested using sample applications for all relevant platforms (iOS/android/flutter/Javascript) that use the feature added/modified

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #592, opened 2023-05-10
Release Amplify Codegen at major version 4

AaronZyLee:
* Revert "fix: codegen downgrade to version 3 (#589)"

This reverts commit c1f9f36979691dfba3dd3db1c4a516aeeb29c41e.

* fix: use hard code codegen version in schema.js

* fix lint

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/appsync-modelgen-plugin/src/visitors/appsync-json-metadata-visitor.ts
+++ b/packages/appsync-modelgen-plugin/src/visitors/appsync-json-metadata-visitor.ts
@@ -160,7 +160,10 @@ export class AppSyncJSONVisitor<
       models: {},
       enums: {},
       nonModels: {},
-      codegenVersion: this._parsedConfig.codegenVersion || 'unknown',
+      // This is hard-coded for the schema version purpose instead of codegen version
+      // To avoid the failure of validation method checkCodegenSchema in JS Datastore
+      // The hard code is starting from amplify codegen major version 4
+      codegenVersion: '3.4.4',
       version: this.computeVersion(),
     };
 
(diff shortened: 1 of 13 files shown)


packages/appsync-modelgen-plugin/src/visitors/appsync-json-metadata-visitor.ts
```diff
@@ -160,7 +160,10 @@ export class AppSyncJSONVisitor<
       models: {},
       enums: {},
       nonModels: {},
-      codegenVersion: this._parsedConfig.codegenVersion || 'unknown',
+      // This is hard-coded for the schema version purpose instead of codegen version
+      // To avoid the failure of validation method checkCodegenSchema in JS Datastore
+      // The hard code is starting from amplify codegen major version 4
```
svidgen, line 165: It might be worth explaining why/when this **should** be revved. I.e.,

> This value should be revved using semver rules when any protocol details expressed in the `/models` files change.

svidgen: Maybe more to the point, this comment sounds like, "we're setting this to X because DataStore breaks otherwise." But, the value actually just identifies a protocol version, and that protocol version now deviates from package version as of package version 4.

AaronZyLee: Yes this will be the protocol version instead. The current approach is a temporary workaround. I will mark this as a TODO to update the comments once the protocol between codegen and library is well defined after a detailed design

phani-srikar (approved):

PR #593, opened 2023-05-11
chore(amplify-codegen): dart api model .fromJson() refactor

Equartey:
#### Description of changes
Flatens the nested data structure of `serializedData` found in the `.fromJson()` method on codegen models. This PR is to be paired with changes in [amplify-flutter#4665](https://github.com/aws-amplify/amplify-flutter/pull/4665)

#### Issue #, if available
https://github.com/aws-amplify/amplify-flutter/issues/816

Change were validated via: 

1. amplify-codegen tests
2. amplify-flutter api category unit & integ tests 
3. amplify-flutter datastore category unit & integ tests 
4. Manual testing within both categories on ios & android devices

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/appsync-modelgen-plugin/src/visitors/appsync-dart-visitor.ts
+++ b/packages/appsync-modelgen-plugin/src/visitors/appsync-dart-visitor.ts
@@ -698,25 +698,26 @@ export class AppSyncModelDartVisitor<
           if (this.isModelType(field)) {
             if (field.isList) {
               return [
-                `${fieldName} = json['${varName}'] is List`,
-                indent(`? (json['${varName}'] as List)`),
-                indent(`.where((e) => e?['serializedData'] != null)`, 2),
-                indent(
-                  `.map((e) => ${this.getNativeType({
-                    ...field,
-                    isList: false,
-                  })}.fromJson(new Map<String, dynamic>.from(e['serializedData'])))`,
-                  2,
-                ),
-                indent(`.toList()`, 2),
+                `${fieldName} = json['${varName}']  != null`,
+                indent(`? json['${varName}'] is Map`),
+                indent(`? (json['${varName}']['items'] as List)`, 2),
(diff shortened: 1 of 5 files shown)


dpilch (dismissed):

packages/appsync-modelgen-plugin/src/visitors/appsync-dart-visitor.ts
```diff
@@ -698,25 +698,26 @@ export class AppSyncModelDartVisitor<
...
-                  2,
-                ),
-                indent(`.toList()`, 2),
+                `${fieldName} = json['${varName}']  != null`,
+                indent(`? json['${varName}'] is Map`),
+                indent(`? (json['${varName}']['items'] as List)`, 2),
+                indent(`.where((e) => e != null)`, 4),
+                indent(`.map((e) => ${this.getNativeType({ ...field, isList: false })}.fromJson(new Map<String, dynamic>.from(e)))`, 4),
+                indent(`.toList()`, 4),
+                indent(`: (json['${varName}'] as List)`),
```
Jordan-Nelson, line 707: Q: Can we be sure that `json['${varName}']` is a List at this point?

The existing logic only casts to a List after checking that the type is List. The new logic casts to a List after checking that it is not null and not a Map. If `json['${varName}']` were ever something other than Map, List, or null this would throw an exception.

The logic in the PR is:
- If map, handle map
- Otherwise, if not null, cast to List and handle list
- Otherwise return null 

Would it make more sense to have:
- If map, handle map
- Otherwise, if List, handle list
- Otherwise, return null

Equartey: Good call out. I adjusted the logic to prefer type checks over null checks.

Jordan-Nelson (approved):

dpilch (approved):

PR #594, opened 2023-05-15
feat: add codegen config related apis

phani-srikar:
#### Description of changes

Adds new API to read parts of the codegen configuration like the path to generated GraphQL documents and frontend types.
[Click to view the API review doc](https://quip-amazon.com/Zsf4AUVFyQlY/API-review-Amplify-Codegen-Configuration-API)
This will enable the customers to programmatically render the imports for generated GraphQL documents and types in their application.

Added unit test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-codegen/src/codegen-config/AmplifyCodeGenConfig.js
```diff
@@ -1,25 +1,24 @@
...
         if (!withoutInit) {
-          projectRoot = amplify.getEnvInfo().projectPath || process.cwd();
+          projectRoot = projectPath || process.cwd();
         } else {
           projectRoot = process.cwd();
         }
-        const configPath = join(projectRoot, '.graphqlconfig.yml');
-        this.gqlConfig = new graphQLConfig.GraphQLConfig(null, configPath);
-        this.gqlConfig.config = {};
+        this.gqlConfig = graphQLConfig.getGraphQLConfig(projectRoot);
```
AaronZyLee, line 21: Not very familiar with the `graphql-config`. Does this change of method call have the same effect as the previous one?

packages/amplify-codegen/src/codegen-config/index.js
```diff
@@ -1,12 +1,69 @@
...
+    const projectPath = context.amplify.getEnvInfo().projectPath;
+    config = new AmplifyCodeGenConfig(projectPath, withoutInit);
   }
   return config;
 }
 
-module.exports = loadConfig;
+function getCodegenConfig(projectPath) {
+  if (!projectPath) {
+    throw new Error('Invalid projectPath provided to get codegen configuration');
```
AaronZyLee, line 17: nit: The `project path` can be used instead of `projectPath` for the error string part. Same as below.

packages/amplify-codegen-e2e-core/src/init/initProjectHelper.ts
```diff
@@ -108,8 +108,7 @@ export function initAndroidProjectWithProfile(cwd: string, settings: Object): Pr
       .wait('Choose your default editor:')
       .sendLine(s.editor)
       .wait("Choose the type of app that you're building")
-      .send('j')
-      .sendCarriageReturn()
+      .sendLine('android')
```
AaronZyLee, line 111: Is this related to the CLI version update?

packages/amplify-codegen/src/codegen-config/index.js
```diff
@@ -1,12 +1,69 @@
...
+  if (!projects.length > 0) {
+    throw new Error(`No projects were found in fetched codegen configuration: ${codegenConfig}`);
+  }
+
+  const cfg = projects[0];
+  const includeFiles = cfg.includes[0];
+
+  const opsGenDirectory = cfg.amplifyExtension.docsFilePath
+    ? cfg.amplifyExtension.docsFilePath
+    : path.dirname(path.dirname(includeFiles));
```
AaronZyLee, line 43: Just curious on the case of docFilePath not being defined. Do we have a walkthrough case ending up in this situation? I look into the [code](https://github.com/aws-amplify/amplify-codegen/blob/f96e002435cc1c4c31d199cd4ffdca17f0856f59/packages/amplify-codegen/src/walkthrough/configure.js#L58) and do not find any unless the `.graphqlconfig` is manually edited.

AaronZyLee: If it happens on manual edit only, I think we also need to check the `includeFiles` is in the correct form of glob expression ending up with `**/.*graphql`

AaronZyLee (changes requested):

packages/amplify-codegen-e2e-core/src/init/initProjectHelper.ts
phani-srikar: Yeah I think so. Bunch of e2es started failing post release, have fixed most.

packages/amplify-codegen/src/codegen-config/index.js
phani-srikar: I kind of took the same logic we use for [statements generation](https://github.com/aws-amplify/amplify-codegen/blob/f96e002435cc1c4c31d199cd4ffdca17f0856f59/packages/amplify-codegen/src/commands/statements.js#L39) to compute the destination.

AaronZyLee (approved):

dpilch (approved):

PR #607, opened 2023-06-16
Fix add codegen from a non-amplify application

phani-srikar:
#### Description of changes

Check if the project is initialized with Amplify before calling the `getEnvInfo` on the shared Amplify context.
This fixes the issue with running `add codegen --apiId <appsync-api-id>` command. The `apiId` input is optional.

Validated in non-amplify and amplify initialized JS applications.
Added e2e test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-codegen-e2e-core/src/categories/codegen.ts
+++ b/packages/amplify-codegen-e2e-core/src/categories/codegen.ts
@@ -31,11 +31,7 @@ export function generateStatementsAndTypes(cwd: string) : Promise<void> {
 // CLI workflow to add codegen to Amplify project
 export function addCodegen(cwd: string, settings: any = {}): Promise<void> {
   return new Promise((resolve, reject) => {
-    const cmdOptions = ['codegen', 'add'];
-    if (settings.withoutInit) {
-      cmdOptions.push('--apiId', 'mockapiid');
-    }
-    const chain = spawn(getCLIPath(), cmdOptions, { cwd, stripColors: true });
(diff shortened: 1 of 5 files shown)


alharris-at (approved):

packages/amplify-codegen-e2e-core/src/categories/codegen.ts
```diff
@@ -31,7 +31,11 @@ export function generateStatementsAndTypes(cwd: string) : Promise<void> {
 // CLI workflow to add codegen to Amplify project
 export function addCodegen(cwd: string, settings: any = {}): Promise<void> {
   return new Promise((resolve, reject) => {
-    const chain = spawn(getCLIPath(), ['codegen', 'add'], { cwd, stripColors: true });
+    const cmdOptions = ['codegen', 'add'];
+    if (settings.withoutInit) {
```
AaronZyLee, line 35: I suggest adding this command into a new method called `addCodegenWithApiId` instead of depending on the flag value as they are two use cases. Also could you confirm that running `amplify codegen add` without init is not allowed(I remember the previous command will give a msg like "please use the apiId instead"?)

phani-srikar: I couldn't find reference to non-amplify workflows in the latest CLI docs but looking at this legacy page, the only requirement to is to have a local introspection schema and with or without apiId should work. https://docs.amplify.aws/cli-legacy/graphql-transformer/codegen/#amplify-add-codegen

alharris-at (approved):

AaronZyLee (approved):

PR #619, opened 2023-07-12
fix: Swift - Generate schema for non-models containing models correctly

MorFix:
#### Description of changes
If a schema contains non-models types that contain models, they're not generated correctly in Swift, because of a missing label, and the code doesn't compile.
That scenario can happen if you have a custom resolver that returns a model along with other things.

#### Issue #, if available
-N/A-

- Added test
- Generated models locally

#### Checklist
- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are [changed or added](https://github.com/aws-amplify/amplify-codegen/blob/main/CONTRIBUTING.md#tests)
- [X] Changes are tested using sample applications for all relevant platforms (iOS/android/flutter/Javascript) that use the feature added/modified

### DIFF
--- a/packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-swift-visitor.test.ts.snap
+++ b/packages/appsync-modelgen-plugin/src/__tests__/visitors/__snapshots__/appsync-swift-visitor.test.ts.snap
@@ -1290,3 +1290,45 @@ extension ModelPath where ModelType == task {
     }
 }"
 `;
+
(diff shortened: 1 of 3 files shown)


dpilch:
Thanks for the contribution @MorFix. Could you also add this test schema here: https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-graphql-schema-test-library/src/index.ts#L39

This will add this schema to our E2E tests that will test for successful compilation of the generated models.

PR #620, opened 2023-07-12
feat: add types to graphql in TS outputs

svidgen:
## Let's discuss the general approach before we worry about the nits.

In order to add types to the generated graphql for TypeScript, we need to know the underlying operation type (`Query | Mutation | Subscription`) and specific operation name (`CreateBlog`, `ListBlogs`, etc.) at the time graphql operations are embeded/formatted for the language as variable assignments. This proposal infers the specific operation name from the key value in the `statements` map.

The alternative would be to weave original type names through the chain &mdash; this seems more robust. But, if the type name is effectively captured in the `statements` map anyway, perhaps this is unnecessary.

I'm looking for high level guidance from codegen folks before proceeding.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-codegen/src/utils/GraphQLStatementsFormatter.js
+++ b/packages/amplify-codegen/src/utils/GraphQLStatementsFormatter.js
@@ -7,10 +7,28 @@ const LINE_DELIMITOR = '\n';
  * Utility class to format the generated GraphQL statements based on frontend language type
  */
 class GraphQLStatementsFormatter {
-  constructor(language) {
+  constructor(language, op) {
     this.language = language || 'graphql';
+    this.opTypeName = {
+      queries: 'Query',
+      mutations: 'Mutation',
+      subscriptions: 'Subscription',
+    }[op];
     this.lintOverrides = [];
     this.headerComments = [];
+  }
+
+  get typeDefs() {
+    if (this.language === 'typescript' && this.opTypeName) {
+      return [
+        `import * as APITypes from '../API';`,
+        `type Generated${this.opTypeName}<InputType, OutputType> = string & {`,
+        `  __generated${this.opTypeName}Input: InputType;`,
+        `  __generated${this.opTypeName}Output: OutputType;`,
+        `};`,
+      ].join(LINE_DELIMITOR);
+    }
+    return '';
   }
 
   format(statements) {
@@ -21,10 +39,7 @@ class GraphQLStatementsFormatter {
         return this.prettify(this.formatJS(statements));
       case 'typescript':
         this.headerComments.push(CODEGEN_WARNING);
-        this.lintOverrides.push(...[
-          '/* tslint:disable */',
-          '/* eslint-disable */'
-        ]);
+        this.lintOverrides.push(...['/* tslint:disable */', '/* eslint-disable */']);
         return this.prettify(this.formatJS(statements));
       case 'flow':
         this.headerComments.push('@flow', CODEGEN_WARNING);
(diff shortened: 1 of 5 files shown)


packages/amplify-codegen/src/utils/GraphQLStatementsFormatter.js
```diff
@@ -7,12 +7,30 @@ const LINE_DELIMITOR = '\n';
...
+      subscriptions: 'Subscription',
+    }[op];
     this.lintOverrides = [];
     this.headerComments = [];
   }
 
+  get typeDefs() {
+    if (this.language === 'typescript' && this.opTypeName) {
+      return [
+        `import * as APITypes from '../API';`,
```
svidgen, line 24: Is the location of `API.ts` stable relative to the generated graphql path? I'm guessing *not*. How should I calculate the correct relative path here?

packages/amplify-codegen/src/utils/GraphQLStatementsFormatter.js
```diff
@@ -36,27 +51,42 @@ class GraphQLStatementsFormatter {
...
+
+    const titleCasedName = `${name[0].toUpperCase()}${name.slice(1)}`;
+    const variablesTypeName = `APITypes.${titleCasedName}${this.opTypeName}Variables`;
+    const resultTypeName = `APITypes.${titleCasedName}${this.opTypeName}`;
+
+    return ` as Generated${this.opTypeName}<
+      ${variablesTypeName},
+      ${resultTypeName}
+    >;`;
+  }
```
svidgen, line 88: I was initially thinking this was *very* gross. But, now I'm wondering if the `name` I receive here (from the `statements` map) is actually all the info I need, minus the casing.

packages/amplify-codegen/src/utils/GraphQLStatementsFormatter.js
```diff
@@ -36,27 +51,42 @@ class GraphQLStatementsFormatter {
...
     return formattedOutput;
   }
 
+  buildTypeTag(name) {
+    if (!this.opTypeName) return '';
+    if (this.language !== 'typescript') return '';
+
+    const titleCasedName = `${name[0].toUpperCase()}${name.slice(1)}`;
+    const variablesTypeName = `APITypes.${titleCasedName}${this.opTypeName}Variables`;
+    const resultTypeName = `APITypes.${titleCasedName}${this.opTypeName}`;
```
svidgen, line 82: I'll have a rev by standup tomorrow that uses `interfaceNameFromOperation` + a new `interfaceVariablesNameFromOperation` instead of re-deriving those names here. If I'm reading things correctly, that should guard against future mismatches due to drift.

packages/amplify-codegen/src/utils/GraphQLStatementsFormatter.js
```diff
@@ -36,27 +51,42 @@ class GraphQLStatementsFormatter {
...
+      LINE_DELIMITOR,
+    );
     return formattedOutput;
   }
 
+  buildTypeTag(name) {
+    if (!this.opTypeName) return '';
+    if (this.language !== 'typescript') return '';
+
+    const titleCasedName = `${name[0].toUpperCase()}${name.slice(1)}`;
```
svidgen, line 80: If I'm reading things right, what I'm deriving here is actually `name` from the `ops` arrays earlier in the chain. It *looks* safe to use this name and just fix casing, but in case I'm wrong and to prevent drift, I'm looking at adding an optional flag to `generateGraphQLDocuments` to return something like `Map<string, {query: string, meta: { ...}}>` when needed.

packages/amplify-codegen/src/utils/GraphQLStatementsFormatter.js
svidgen: Added as another PR to compare/contrast: https://github.com/aws-amplify/amplify-codegen/pull/623/files

packages/amplify-codegen/src/utils/GraphQLStatementsFormatter.js
svidgen: Added as another PR to compare/contrast: https://github.com/aws-amplify/amplify-codegen/pull/623/files

packages/amplify-codegen/src/utils/GraphQLStatementsFormatter.js
phani-srikar: No it is not, customers can choose any custom path relative to the project root when they do "amplify add codegen".

svidgen:
Using this one instead: https://github.com/aws-amplify/amplify-codegen/pull/623

PR #630, opened 2023-07-14
feat: API.swift compatibility with AWSAPIPlugin

lawmicha:
#### Description of changes
Currently developers that have an AppSync backend that was not provisioned with Amplify CLI can still generate the **API.swift** file from the introspection schema. This file has a hard dependency on [iOS AppSync SDK client.](https://github.com/awslabs/aws-mobile-appsync-sdk-ios/). The top of API.swift contains `imports AWSAppSync`.

This file is pretty useful as it contains all the different operations available and  components to successfully make a request, but cannot be used with Amplify's APIPlugin. This PR makes the API.swift file compatible with Amplify AWSAPIPlugin, removing the dependency on AppSync SDK.
1. If can import AWSAPIPlugin, declare the types required to compile API.swift in-line. The required types were extracted out of AWSAppSync SDK codebase. Additionally, the Data types were made decodable, to allow them to be passed to Amplify.GraphQLRequest as the type to decode to.
2. Fallback to importing AWSAppSync. This is to allow backwards compatibility with existing app clients. Developers may continue to their existing workflow in updating their schema, and regenerate the API.swift file, while not having to change their code to use AWSAPIPlugin.

We considered that if the developer has AppSync SDK and Amplify already in the same app, and **API.swift** was compilng due to AppSync SDK. The developer is most likely using AppSync SDK to interact with the backend, while the rest of Amplify (excluding AWSAPIPlugin) for the remaining plugins. If they did not add AWSAPIPlugin, there's no impact and they can upgrade to use AWSAPIPlugin when they are ready to.

if AWSAPIPlugin was added, it may have been
1. Added accidentally and unused, developer may have to remove the AWSPIPlugin so it continues to compile with AWSAppSync import, or upgrade to AWSAPIPlugin call pattern in their code.
2. The developer is using AWSAPIPlugin's REST APIs and AWSAppSync SDK. We believe the number of customers in this use case is near zero. Because we think this is an edge case, the path to unblock the customer is that they will have to upgrade to use AWSAPIPlugin, or modify the API.swift to reverse the changes to import AWSAppSync instead.
3. The developer is using AWSAPIPlugin's GraphQL requests already, and included AppSync SDK to compile the API.swift successfully. The developer can now remove AppSync SDK as it is an unused dependency. 

The upgrade guide is meant for both existing customers to move to AWSAPIPlugin and new customers that have built AppSync backends without using the Amplify's provisioning, see upgrade guide PR for more details: https://github.com/aws-amplify/docs/pull/5649

Local testing:
```
nvm use v14 && yarn clean && rm -rf node_modules && yarn && yarn build && yarn test
```

`amplify-dev` testing
```
nvm use 16.14.0 && yarn setup-dev
```

We validate the new API.swift file compiles and works during runtime by the integration tests in the following PRs:
- https://github.com/aws-amplify/amplify-swift/pull/3098

### Tagged release

`@aws-amplify/cli@12.2.1-swift-codegen.0`

example of CLI required changes to create CLI tagged release: https://github.com/aws-amplify/amplify-cli/commit/3e4d27a8620e51d322ea572da05a265bf6aab470

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/graphql-types-generator/src/swift/appSyncCompatibilityTypes.ts
+++ b/packages/graphql-types-generator/src/swift/appSyncCompatibilityTypes.ts
@@ -48,7 +48,7 @@ public extension GraphQLMapConvertible {
 
 public typealias GraphQLID = String
 
-public protocol GraphQLOperation: AnyObject {
+public protocol APISwiftGraphQLOperation: AnyObject {
   
   static var operationString: String { get }
   static var requestString: String { get }
@@ -59,7 +59,7 @@ public protocol GraphQLOperation: AnyObject {
   associatedtype Data: GraphQLSelectionSet
 }
 
-public extension GraphQLOperation {
+public extension APISwiftGraphQLOperation {
   static var requestString: String {
     return operationString
   }
@@ -73,11 +73,11 @@ public extension GraphQLOperation {
   }
 }
 
-public protocol GraphQLQuery: GraphQLOperation {}
+public protocol GraphQLQuery: APISwiftGraphQLOperation {}
 
-public protocol GraphQLMutation: GraphQLOperation {}
+public protocol GraphQLMutation: APISwiftGraphQLOperation {}
 
-public protocol GraphQLSubscription: GraphQLOperation {}
+public protocol GraphQLSubscription: APISwiftGraphQLOperation {}
 
 public protocol GraphQLFragment: GraphQLSelectionSet {
   static var possibleTypes: [String] { get }
--- a/packages/graphql-types-generator/src/swift/appSyncCompatibilityTypes.ts
+++ b/packages/graphql-types-generator/src/swift/appSyncCompatibilityTypes.ts
@@ -1,0 +1,419 @@
+// Blank lines at the beginning are intentional
+export const appSyncCompatibilityTypesCode = `
+import Foundation
+
+public protocol GraphQLInputValue {
+}
+
+public struct GraphQLVariable {
+  let name: String
+  
+  public init(_ name: String) {
+    self.name = name
+  }
+}
+
+extension GraphQLVariable: GraphQLInputValue {
+}
+
+extension JSONEncodable {
+  public func evaluate(with variables: [String: JSONEncodable]?) throws -> Any {
+    return jsonValue
+  }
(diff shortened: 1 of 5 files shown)


packages/graphql-types-generator/src/swift/appSyncCompatibilityTypes.ts
```diff
@@ -0,0 +1,550 @@
...
+        if let jsonObject = try? JSONValue(from: decoder) {
+            let encoder = JSONEncoder()
+            let jsonData = try encoder.encode(jsonObject)
+            let decodedDictionary = try JSONSerialization.jsonObject(with: jsonData, options: []) as! [String: Any]
+            let optionalDictionary = decodedDictionary.mapValues { $0 as Any? }
+
+            self.init(snapshot: optionalDictionary)
+        } else {
+            self.init(snapshot: [:])
+        }
```
lawmicha, line 223: This is where the custom decoder was added. It uses a self defined JSONValue type to get the data into an JSONValue object before serializing and deserializing back to the expected output a `Snapshot` (typealias for [String:Any?])

AaronZyLee (dismissed):

packages/graphql-types-generator/src/swift/appSyncCompatibilityTypes.ts
```diff
@@ -0,0 +1,419 @@
...
+    }
+}
```
lawmicha, line 108: This is where the custom decoder was added. It uses an internally defined JSON value type called `APISwiftJSONValue` (to avoid developer defined "JSONValue" type) to get the data into an JSON object before serializing and deserializing back to the expected output a Snapshot (typealias for [String:Any?]))

dpilch:
Are there any unit tests covering this change?

5d (approved):

lawmicha:
> Are there any unit tests covering this change?

good point, @dpilch just added one to verify the `fileHeader` function!

dpilch (approved):

5d (approved):

dpilch (approved):

PR #651, opened 2023-07-25
fix: post publish sync between main and release

phani-srikar:
#### Description of changes

The codebuild job starts with HEAD at a detached commit from the `release` branch, which is observed in the SSH session. Another thing is that there is no default tracking branches for the cloned repo in the container.
Adding explicit `pull` from origin for `main` and `release` to be sure we're using latest code and the history is all fetched.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/shared-scripts.sh
+++ b/shared-scripts.sh
@@ -319,7 +319,6 @@ function runE2eTest {
 
 function _deploy {
   echo "Deploy"
-  loadCacheFromBuildJob
   echo "Authenticate with NPM"
   PUBLISH_TOKEN=$(echo "$NPM_PUBLISH_TOKEN" | jq -r '.token')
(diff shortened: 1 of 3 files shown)


alharris-at (approved):

shared-scripts.sh
```diff
@@ -319,7 +319,6 @@ function runE2eTest {
 
 function _deploy {
   echo "Deploy"
-  loadCacheFromBuildJob
```
AaronZyLee, line 322: Is this removed because there is an additional `yarn production-build` running in deploy?

AaronZyLee (approved):

shared-scripts.sh
phani-srikar: Yes we probably don't need it. Tag release atleast works without it.

PR #655, opened 2023-07-29
feat: some ops improvements

phani-srikar:
#### Description of changes

- Ability to view multiple E2E test artifacts
- Ability to SSH into multiple failing E2E tests using a single `yarn cloud-e2e-debug <batch_id>` command and not requiring us to commit the `debug_workflow` buildspec file.
- Ability to run release workflow from any branch and fail fast for tag release workflow if there are un-pushed local commits
- Remove lint job from release workflow since it's already run on every PR and is non-blocking

Tested them locally and the full E2E run succeeded.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


scripts/view-test-artifacts.ts
```diff
@@ -0,0 +1,189 @@
...
+  buildStatus: BuildStatus;
+  artifactLocation?: string;
+};
+
+/**
+ * Utility method to filter out incomplete test artifact definitions.
+ * @param artifact the potentially incomplete artifact to validate
+ * @returns whether the artifact is complete or not
+ */
+const testArtifactIsComplete = (artifact: Partial<TestArtifact>): artifact is TestArtifact => artifact.jobName !== undefined;
```
dpilch, line 47: Shouldn't this be based on `buildStatus` not `jobName`?

phani-srikar: This check is intended towards filtering the tests without an identifier since Codebuild SDK call for getting batches could return `undefined` job_names. We have subsequent filtering for build status in place, so the in_progress ones are shown as such on the UI.
Screenshot 2023-07-31 at 12 55 44 PM

AaronZyLee (approved):
Is it good to keep the breakpoint and debug session if we trigger the build without `debug-session-enabled` flag? (passing through if flag is undefined)

phani-srikar:
> Is it good to keep the breakpoint and debug session if we trigger the build without `debug-session-enabled` flag? (passing through if flag is undefined)

It's a no-op and skips breakpoints if we don't trigger the build with `enable-debug-session`.

.codebuild/run_ios_modelgen_e2e_test.yml
```diff
@@ -10,7 +10,12 @@ env:
 phases:
   build:
     commands:
+      - source ./shared-scripts.sh && _setupE2ETestsLinux
+      - codebuild-breakpoint
       - source ./shared-scripts.sh && _runE2ETestsLinux
+      - unset AWS_ACCESS_KEY_ID
+      - unset AWS_SECRET_ACCESS_KEY
+      - unset AWS_SESSION_TOKEN
```
phani-srikar, line 18: We need to unset these to fall back to codebuild role from the E2E parent role before we populate the generated models in S3 artifacts bucket.

alharris-at (approved):

AaronZyLee (approved):

PR #670, opened 2023-08-21
refactor: create graphql-generator entry point for types

dpilch:
#### Description of changes

* Add `generateTypes` to graphql-generator
* Refactor amplify-codegen to use `generateTypes`

#### Issue #, if available

N/A

* Unit testing
* E2E tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/graphql-types-generator/src/generate.ts
```diff
@@ -32,82 +45,133 @@ export default function generate(
...
+  validateQueryDocument(graphqlSchema, document);
+  const output = generateForTarget(graphqlSchema, document, '', target, multipleSwiftFiles, options);
+
+  if (isBasicGeneratedFileMap(output)) {
+    return Object.entries(output)
+      .map(([filepath, file]) => [filepath, file.output])
+      .reduce((acc, [filepath, fileOutput]) => ({ ...acc, [filepath]: fileOutput }), {});
   }
+
+  const filename = getOutputFileName('', target);
```
alharris-at, line 103: Why is the input filename hardcoded as an empty string?

alharris-at: Looks like for classic use-cases the fileName will be imputed from the target, and this is added to support override paths/names, is that right?

dpilch: It is an empty string so that this logic for angular will run.
https://github.com/aws-amplify/amplify-codegen/pull/670/files#diff-e85128fcfb85eaa6cb2036bdacd19d627e7bad0816c86d5fa931a23b248f1348R6

```typescript
inputFileName = inputFileName || (target === 'angular' ? 'api' : 'API');
```

Probably should `null`/`undefined`. I'll clean this up a little.

alharris-at: Discussed offline. the result filename for JS targets is actually inconsequential in the new API surface area, so this is basically a no-op. Leaving as-is to keep from impacting existing use-cases.

alharris-at (approved):

PR #694, opened 2023-09-12
feat: support modelgen without an amplify backend

alharris-at:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-codegen/src/utils/getProjectRoot.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * Find the project root.
+ * @param {!import('@aws-amplify/amplify-cli-core').$TSContext} context the amplify runtime context
+ * @returns {!string} the project root, or cwd
+ */
+const getProjectRoot = (context) => {
```
alharris-at, line 6: Moved out of `commands/model.js` since it's reused in the `getOutputDirParam` utility.

packages/amplify-codegen/src/commands/models.js
```diff
@@ -34,45 +36,55 @@ const modelgenFrontendToTargetMap = {
...
 /**
  * Returns feature flag value, default to `false`
- * @param {!string} key feature flag id
+ * @param {!AmplifyContext} context the amplify runtime context
+ * @param {!string} flagName feature flag id
  * @returns {!boolean} the feature flag value
  */
-const readFeatureFlag = key => {
-  let flagValue = false;
+const readFeatureFlag = (context, flagName) => {
```
dpilch, line 53: With this defaulting to false, will every feature flag need to be passed? 

i.e. Outside of amplify project. `amplify codegen models --schema-path schema.graphql --target typescript`. This will have all feature flags as false?

I think if we are in a no init state we should check if the feature flag was passed in the CLI. If it is not passed we should return `undefined` and have the graphql-generator package handle the defaults for each feature flag.

alharris-at: It only defaults to false if the call to FeatureFlags throws, so we'll still receive the 'New Project Default' value for each flag.

dpilch (approved):

PR #700, opened 2023-09-14
fix: avoid name collision on java model build step

dpilch:
#### Description of changes

Avoid name collision on Java model build step by adding `Build` if name is used by a type.

See https://github.com/aws-amplify/amplify-codegen/issues/695 for details.

It is still possible for a collision to occur if the model has a second type that includes `Build`.

```
type TutorialStep {
  position: Int!
  text: String!
}

type TutorialBuild {
  position: Int!
  text: String!
}

type MyObject @model {
  id: ID!
  tutorial: [TutorialStep]
  tutorialBuild: [TutorialBuildStep]
}
```

We could add another layer of checks or use a randomly generated string, but I think this case is so rare that it should be ok.

#### Issue #, if available

Resolves https://github.com/aws-amplify/amplify-codegen/issues/695

Unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/appsync-modelgen-plugin/src/visitors/appsync-java-visitor.ts
+++ b/packages/appsync-modelgen-plugin/src/visitors/appsync-java-visitor.ts
@@ -13,7 +13,14 @@ import {
 } from '../configs/java-config';
 import { JAVA_TYPE_IMPORT_MAP } from '../scalars';
 import { JavaDeclarationBlock } from '../languages/java-declaration-block';
-import { AppSyncModelVisitor, CodeGenField, CodeGenModel, CodeGenPrimaryKeyType, ParsedAppSyncModelConfig, RawAppSyncModelConfig } from './appsync-visitor';
+import {
+  AppSyncModelVisitor,
+  CodeGenField,
+  CodeGenModel,
+  CodeGenPrimaryKeyType,
+  ParsedAppSyncModelConfig,
+  RawAppSyncModelConfig,
+} from './appsync-visitor';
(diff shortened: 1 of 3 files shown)


packages/appsync-modelgen-plugin/src/visitors/appsync-java-visitor.ts
```diff
@@ -843,10 +890,27 @@ export class AppSyncModelJavaVisitor<
...
    * @returns string
    */
-  private getStepInterfaceName(nextFieldName: string): string {
-    return `${pascalCase(nextFieldName)}Step`;
+  private getStepInterfaceName(nextFieldName: string, types: Set<string>): string {
+    const pascalCaseFieldName = pascalCase(nextFieldName);
+    const stepInterfaceName = `${pascalCaseFieldName}Step`;
+
+    if (types.has(stepInterfaceName)) {
+      return `${pascalCaseFieldName}BuildStep`;
```
AaronZyLee, line 901: This could still collide if for a model `A`, there are `AStep` and `ABuildStep` in the fields. But it should be a rare case. The implementation is enough to fix the issue.

AaronZyLee (approved):

alharris-at (approved):

packages/appsync-modelgen-plugin/src/visitors/appsync-java-visitor.ts
dpilch: I agree that this case would be so rare that this fix should be sufficient.

PR #718, opened 2023-09-26
fix: generate multiple swift files

dpilch:
#### Description of changes

Generate multiple Swift type files. The separate files are for each GraphQL statement file. The multiple type files need the names of the GraphQL statements files in order to generate with the correct names.

This leaves us a few options for the queries arg.
1. `[fileContents: string, filename: string][]`
2. `Source[]` (from `graphql` package).
3. Type union of the current type (string) and one of the two above options.

I chose `string | Source[]` for this PR, but I don't have strong feelings on the end direction.

#### Issue #, if available

https://github.com/aws-amplify/amplify-codegen/issues/711

* Unit tests
* Manual tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-codegen/src/commands/types.js
+++ b/packages/amplify-codegen/src/commands/types.js
@@ -2,12 +2,13 @@ const path = require('path');
 const fs = require('fs-extra');
 const Ora = require('ora');
 const glob = require('glob-all');
+const { Source } = require('graphql');
 
 const constants = require('../constants');
 const { loadConfig } = require('../codegen-config');
 const { ensureIntrospectionSchema, getFrontEndHandler, getAppSyncAPIDetails, getAppSyncAPIInfoFromProject } = require('../utils');
 const { generateTypes: generateTypesHelper } = require('@aws-amplify/graphql-generator');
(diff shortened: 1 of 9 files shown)


packages/amplify-codegen/src/commands/types.js
```diff
@@ -93,32 +93,25 @@ async function generateTypes(context, forceDownloadSchema, withoutInit = false,
         codeGenSpinner.start();
         const schema = fs.readFileSync(schemaPath, 'utf8');
         const introspection = path.extname(schemaPath) === '.json';
-
+        const multipleSwiftFiles = target === 'swift' && fs.existsSync(outputPath) && fs.statSync(outputPath).isDirectory();
```
alharris-at, line 96: If I'm running for the first time will `outputPath` necessarily exist? Do we instead need to rely on the path filename suffix or something like that?

alharris-at: Ahh, I just saw that's the old behavior as well.

alharris-at (approved):

phani-srikar (approved):

Issue #734, opened 2023-10-04
Amplify codegen fails with error:  "Cannot use 'in' operator to search for 'name' in null"


ErlendHer:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-codegen/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-codegen/blob/main/CONTRIBUTING.md#bugs).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

npm

### If applicable, what version of Node.js are you using?

v18.17.1

### Amplify CLI Version

12.5.2

### What operating system are you using?

MacOS Sonoma

### Amplify Codegen Command

codegen, codegen types

### Describe the bug

When running `amplify codegen`. We're getting this error: 

`Cannot use 'in' operator to search for 'name' in null`

And `API.ts` fails to generate. However, the `mutations.ts`, `queries.ts` and `subscriptions.ts` all compile.

I'm guessing there is an "in" check somewhere in the code on a null value where an arary / object was expected. That's why we like null-safety and TypeScript <3 (below code produces the same error)
```TypeScript
const iCouldBeNull = null as any;
console.log('name' in iCouldBeNull);
```

Our schema compiles without an error.

This happened after upgrading the CLI from 12.1.1 to 12.5.2.

I tried downgrading the CLI and our schema compiles as expected.

As a side note, when using the CLI 12.5.2, we're also getting errors due to the imports not having `.js` extensions, and since we're using ESM and not CommonJS, this doesn't work for our project and we have to update imports manually with the new version.

### Expected behavior

Amplify codegen should work with a valid schema and provide the correct generated types.

### Reproduction steps

This issue only occurs if there are files in the otuput directory before running amplify codegen.

1. `amplify api gql-compile`
2. `amplify codegen`
3. `amplify codegen`

The third step should fail with the error.

Running the same code with CLI version 12.1.1 (version we used before) works without issues.

### GraphQL schema(s)

<details>

```graphql
# Put schemas below this line


```

</details>

### Log output

<details>

```
# Put your logs below this line


```

</details>

ErlendHer:
So we discovered what is causing the issue, and it's pretty insane I must say.

[...]

PR #738, opened 2023-10-11
fix: skip bad statement source on type generation and exclude type file

dpilch:
#### Description of changes

* Skip bad statement source on type generation. If the GraphQL source is `null` remove it from the sources.
* Automatically exclude the generated type file.

#### Issue #, if available

Resolves: https://github.com/aws-amplify/amplify-codegen/issues/734

* Unit testing
* Manual testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-codegen/src/commands/types.js
+++ b/packages/amplify-codegen/src/commands/types.js
@@ -62,18 +62,27 @@ async function generateTypes(context, forceDownloadSchema, withoutInit = false, 
           cwd: projectPath,
           absolute: true,
         });
-        const queries = queryFilePaths.map(queryFilePath => {
-          const fileContents = fs.readFileSync(queryFilePath, 'utf8');
-          if (
-            queryFilePath.endsWith('.jsx') ||
-            queryFilePath.endsWith('.js') ||
-            queryFilePath.endsWith('.tsx') ||
-            queryFilePath.endsWith('.ts')
-          ) {
-            return extractDocumentFromJavascript(fileContents, '');
-          }
-          return new Source(fileContents, queryFilePath);
-        });
+        const queries = queryFilePaths
+          .map(queryFilePath => {
+            const fileContents = fs.readFileSync(queryFilePath, 'utf8');
+            if (
+              queryFilePath.endsWith('.jsx') ||
+              queryFilePath.endsWith('.js') ||
+              queryFilePath.endsWith('.tsx') ||
+              queryFilePath.endsWith('.ts')
+            ) {
+              return [queryFilePath, extractDocumentFromJavascript(fileContents, '')];
+            }
+            return [queryFilePath, new Source(fileContents, queryFilePath)];
+          })
+          .filter(([queryFilePath, source]) => {
+            if (!source) {
+              context.print.warning(`Unable to extract GraphQL queries from ${queryFilePath}. Skipping source.`);
+              return false;
+            }
+            return true;
+          })
+          .map(([, source]) => source);
         if (queries.length === 0) {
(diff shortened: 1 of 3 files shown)


packages/amplify-codegen/src/commands/types.js
```diff
@@ -58,22 +58,31 @@ async function generateTypes(context, forceDownloadSchema, withoutInit = false,
         const target = cfg.amplifyExtension.codeGenTarget;
 
         const excludes = cfg.excludes.map(pattern => `!${pattern}`);
-        const queryFilePaths = glob.sync([...includeFiles, ...excludes], {
+        const queryFilePaths = glob.sync([...includeFiles, ...excludes, `!${generatedFileName}`], {
```
alharris-at, line 61: Should `generatedFileName` instead be added to the cfg.excludes list during `codegen add/update`? Then I don't think we'd require any other code change, right? Or would we still end up requiring the `Source` filtering down below?

dpilch: Other files that are in the includes glob can lead to the error. i.e. (`src/graphql/foo.ts`). We could say user error at that point, but we should at least improve the error messaging to say to add these files to the excludes.

packages/amplify-codegen/src/commands/types.js
```diff
@@ -62,18 +62,27 @@ async function generateTypes(context, forceDownloadSchema, withoutInit = false,
...
+              queryFilePath.endsWith('.tsx') ||
+              queryFilePath.endsWith('.ts')
+            ) {
+              return [queryFilePath, extractDocumentFromJavascript(fileContents, '')];
+            }
+            return [queryFilePath, new Source(fileContents, queryFilePath)];
+          })
+          .filter(([queryFilePath, source]) => {
+            if (!source) {
+              context.print.warning(`Unable to extract GraphQL queries from ${queryFilePath}. Skipping source.`);
```
dpilch, line 80: We could also add something like:

```
This file matched the includes target in .grapqhlconfig.yml. Modify the includes or excludes target if this file should not be included.
```

alharris-at: Yeah, I think a CTA here would be helpful, in addition to the error message.

packages/amplify-codegen/tests/commands/types-mock-fs.test.js
```diff
@@ -121,4 +122,21 @@ describe('command - types (mock fs)', () => {
     await generateTypes(MOCK_CONTEXT, false);
     expect(fs.existsSync(generatedFileName)).toBeFalsy();
   });
+
+  it.only('should skip invalid sources', async () => {
```
palpatim, line 126: Is the `only` intentional?

packages/amplify-codegen/tests/walkthrough/configure.test.js
```diff
@@ -78,13 +83,40 @@ describe('configure walk-through', () => {
...
+      amplifyExtension: {
+        graphQLApiId: mockConfigs[1].amplifyExtension.graphQLApiId,
+        generatedFileName: mockGeneratedFileName,
+        codeGenTarget: mockTargetLanguage,
+        maxDepth: MOCK_MAX_DEPTH,
+      },
+    });
+  });
+
+  it('should not generated file to excludes twice', async () => {
```
palpatim, line 104:
```suggestion
  it('should not generate file to excludes twice', async () => {
```

palpatim (approved):

alharris-at (approved):

PR #742, opened 2023-10-13
fix: correctly remove file extension on types import path

dpilch:
#### Description of changes

Correctly remove file extensions on types import path. See https://github.com/aws-amplify/amplify-codegen/issues/741 for details.

#### Issue #, if available

Resolves: https://github.com/aws-amplify/amplify-codegen/issues/741

#### Description of how you validated changes

* Unit testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed or added](https://github.com/aws-amplify/amplify-codegen/blob/main/CONTRIBUTING.md#tests)
- [x] Changes are tested using sample applications for all relevant platforms (iOS/android/flutter/Javascript) that use the feature added/modified

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #743, opened 2023-10-16
ci: run unit test on windows

dpilch:
#### Description of changes

Run unit tests on windows in CodeBuild.

Also fix issues where tests failed on windows.

There is no "codegen core" package to share utilities between all the packages. I used `.split(path.win32.sep).join(path.posix.sep)` wherever the paths needed to be cleaned.

Unit testing on windows

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-codegen/src/codegen-config/AmplifyCodeGenConfig.js
+++ b/packages/amplify-codegen/src/codegen-config/AmplifyCodeGenConfig.js
@@ -2,7 +2,8 @@ const graphQLConfig = require('graphql-config');
 const { isAbsolute, relative, join } = require('path');
 const slash = require('slash');
 const { graphQlToAmplifyConfig } = require('./utils');
-const fs =  require('fs-extra');
+const fs = require('fs-extra');
+const path = require('path');
 
 class AmplifyCodeGenConfig {
   static configFileName = '.graphqlconfig.yml';
@@ -15,10 +16,9 @@ class AmplifyCodeGenConfig {
       if (e instanceof graphQLConfig.ConfigNotFoundError) {
         const projectRoot = projectPath || process.cwd();
         const configPath = join(projectRoot, '.graphqlconfig.yml');
-        if(fs.existsSync(configPath)) {
+        if (fs.existsSync(configPath)) {
           this.gqlConfig = graphQLConfig.getGraphQLConfig(projectRoot);
-        }
-        else {
+        } else {
           this.gqlConfig = new graphQLConfig.GraphQLConfig(null, configPath);
           this.gqlConfig.config = {};
         }
@@ -45,7 +45,9 @@ class AmplifyCodeGenConfig {
     if (!this.constructor.isValidAmplifyProject(project)) {
       return false;
     }
-    const schemaPath = isAbsolute(project.schema) ? relative(this.gqlConfig.configDir, project.schema) : project.schema;
+    const schemaPath = (isAbsolute(project.schema) ? relative(this.gqlConfig.configDir, project.schema) : project.schema)
+      .split(path.win32.sep)
+      .join(path.posix.sep);
     const newProject = {
       schemaPath,
       includes: project.includes,
--- a/packages/amplify-codegen/src/codegen-config/AmplifyCodeGenConfig.js
+++ b/packages/amplify-codegen/src/codegen-config/AmplifyCodeGenConfig.js
@@ -45,6 +45,8 @@ class AmplifyCodeGenConfig {
     if (!this.constructor.isValidAmplifyProject(project)) {
       return false;
     }
+    // Set schemaPath to use posix separators. Node can handle windows and posix separators regradless of platform
+    // Ensures all paths in .graphlqconfig.yml use posix style
     const schemaPath = (isAbsolute(project.schema) ? relative(this.gqlConfig.configDir, project.schema) : project.schema)
       .split(path.win32.sep)
       .join(path.posix.sep);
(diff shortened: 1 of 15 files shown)


packages/amplify-codegen/tests/commands/statements.test.js
```diff
@@ -87,7 +87,7 @@ describe('command - statements', () => {
     await generateStatements(MOCK_CONTEXT, forceDownload);
     expect(getFrontEndHandler).toHaveBeenCalledWith(MOCK_CONTEXT);
     expect(loadConfig).toHaveBeenCalledWith(MOCK_CONTEXT, false);
-    expect(getRelativeTypesPath).toHaveBeenCalledWith('MOCK_PROJECT_ROOT/MOCK_STATEMENTS_PATH', 'API.TS');
+    expect(getRelativeTypesPath).toHaveBeenCalledWith(path.join(MOCK_PROJECT_ROOT, MOCK_STATEMENTS_PATH), 'API.TS');
```
dpilch, line 90: It doesn't matter if `getRelativeTypePath` is called with a path with posix or windows separators.

packages/graphql-generator/src/models.ts
```diff
@@ -61,7 +62,7 @@ export async function generateModels(options: GenerateModelsOptions): Promise<Ge
   return Promise.all(
     appsyncLocalConfig.map(async config => {
       const content = await codegen(config);
-      return { [config.filename]: content };
+      return { [config.filename.split(path.win32.sep).join(path.posix.sep)]: content };
```
dpilch, line 65: This will change the output of `generateModels` so that the filepath will always use posix separators. IMO this should be the correct behavior so that the package is consistent across OS.

packages/graphql-generator/src/utils/GraphQLStatementsFormatter.ts
```diff
@@ -36,10 +36,9 @@ export class GraphQLStatementsFormatter {
     this.lintOverrides = [];
     this.headerComments = [];
     if (typesPath) {
+      const { dir, name } = path.parse(typesPath);
       // ensure posix path separators are used
-      const typesPathWithPosixSep = typesPath.split(path.win32.sep).join(path.posix.sep)
-      const { dir, name } = path.parse(typesPathWithPosixSep);
-      const typesPathWithoutExtension = path.join(dir, name);
+      const typesPathWithoutExtension = path.join(dir, name).split(path.win32.sep).join(path.posix.sep);
```
dpilch, line 41: Fixing the issue introduced by https://github.com/aws-amplify/amplify-codegen/pull/742. Shows how easy it is to mess up without the unit testing in Windows. Thankfully the change has not been released.

packages/graphql-types-generator/src/utilities/getOutputFileName.ts
```diff
@@ -9,7 +9,7 @@ export function getOutputFileName(inputFileName: string, target: Target): string
     const ext = path.extname(inputFileName);
     const baseName = inputFileName.substr(0, inputFileName.length - ext.length);
     const filename = inputFileName.includes(fileExtension) ? inputFileName : `${baseName}.${fileExtension}`;
-    return ['API', 'api'].includes(inputFileName) ? path.join(folderMap[target], filename) : filename;
+    return ['API', 'api'].includes(inputFileName) ? path.join(folderMap[target], filename).split(path.win32.sep).join(path.posix.sep) : filename;
```
dpilch, line 12: This ensure the key for the types file uses posix separators.

packages/amplify-codegen/src/codegen-config/AmplifyCodeGenConfig.js
```diff
@@ -45,7 +45,9 @@ class AmplifyCodeGenConfig {
     if (!this.constructor.isValidAmplifyProject(project)) {
       return false;
     }
-    const schemaPath = isAbsolute(project.schema) ? relative(this.gqlConfig.configDir, project.schema) : project.schema;
+    const schemaPath = (isAbsolute(project.schema) ? relative(this.gqlConfig.configDir, project.schema) : project.schema)
```
dpilch, line 48: This will change the `schemaPath` in `.graphqlconfig.yml` to use posix separators. There will be no functional difference in codegen because node understands paths of both styles.

I think this is the correct behavior because other paths inside `.graphlqconfig.yml` use posix style paths.

packages/amplify-codegen/src/commands/add.js
```diff
@@ -134,7 +134,7 @@ async function add(context, apiId = null, region = 'us-east-1') {
     projectName: withoutInit ? 'Codegen Project' : apiDetails.name,
     includes: answer.includePattern,
     excludes: [...answer.excludePattern, answer.generatedFileName],
-    schema,
+    schema: schema.split(path.win32.sep).join(path.posix.sep),
```
dpilch, line 137: Same behavior as the change in the file directly above.

shared-scripts.sh
```diff
@@ -3,15 +3,29 @@
...
 function storeCache {
   localPath="$1"
   alias="$2"
+  environment="$3"
   s3Path="s3://$CACHE_BUCKET_NAME/$CODEBUILD_SOURCE_VERSION/$alias"
   echo "Writing cache folder $alias to $s3Path"
   # zip contents and upload to s3
-  if ! (cd $localPath && tar cz . | aws s3 cp - $s3Path); then
-      echo "Something went wrong storing the cache folder $alias."
+  errorMessage="Something went wrong storing the cache folder $alias."
```
palpatim, line 18: nit: add "continuing anyway" clause to your error message as below, since it looks like an error storing the cache isn't fatal

dpilch: It isn't fatal, but it probably should be. If we fail to store the cache the test step will fail later. I'll update the error message now, and follow up with a change to fail the build.

palpatim (approved):

alharris-at (approved):

PR #750, opened 2023-10-23
ci: e2e tests on windows

dpilch:
#### Description of changes

Run E2E tests on Windows.

The Windows bash terminal on CodeBuild appears to be non-interactive. This is causing issues with selecting a region when running `amplify configure`. I've set all the Windows E2E tests to run `us-east-1` for now and we can fix that later when you have some more time.

There are two disabled tests for windows only:
1. amplify pull on an ios project.
2. Cypress run after generating graphql statements.

Test 1 is not very valuable. It is possible to do this with the current CLI, but realistically no customer would do this because XCode is not available on windows.

Test 2 is much more valuable. This asserts that the generated graphql statements work in a running app. The app compiles but then there is an error in cypress. It does not appear to be an error in a test case. It might be an issue with cypress on the windows instance.

I was not able to reproduce either case on my own windows workspace. 

To recap, we have 3 follow ups to complete later:
1. Windows E2E tests on other regions
2. Enable amplify pull on ios project test
3. Enable Cypress run after generating graphql statements test.

#### Issue #, if available

N/A

Run the E2E tests.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.codebuild/scripts/e2e_test_windows.sh
+++ b/.codebuild/scripts/e2e_test_windows.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+# set exit on error to true
+set -e
+
+source ./shared-scripts.sh && _setupE2ETestsWindows
+codebuild-breakpoint
+source ./shared-scripts.sh && _runE2ETestsWindows
--- a/.codebuild/scripts/local_publish_helpers.sh
+++ b/.codebuild/scripts/local_publish_helpers.sh
@@ -7,7 +7,7 @@ function startLocalRegistry {
     # Start local registry
     tmp_registry_log="$(mktemp)"
     echo "Registry output file: $tmp_registry_log"
-    (cd && nohup npx ${VERDACCIO_PACKAGE:-$default_verdaccio_package} -c $1 &>$tmp_registry_log &)
+    nohup npx ${VERDACCIO_PACKAGE:-$default_verdaccio_package} -c $1 &>$tmp_registry_log &
     # Wait for Verdaccio to boot
     grep -q 'http address' <(tail -f $tmp_registry_log)
 }
--- a/packages/amplify-codegen-e2e-core/package.json
+++ b/packages/amplify-codegen-e2e-core/package.json
@@ -32,7 +32,7 @@
     "jest-circus": "^27.5.1",
     "jest-environment-node": "^26.6.2",
     "lodash": "^4.17.19",
-    "node-pty": "beta",
+    "node-pty": "1.0.0",
     "retimer": "2.0.0",
     "rimraf": "^3.0.0",
     "strip-ansi": "^6.0.0",
--- a/packages/amplify-codegen-e2e-core/src/categories/api.ts
+++ b/packages/amplify-codegen-e2e-core/src/categories/api.ts
@@ -49,9 +49,7 @@ export function addApiWithoutSchema(cwd: string, opts: Partial<AddApiOptions & {
       .sendCarriageReturn()
       .wait('Do you want to edit the schema now?')
       .sendConfirmNo()
-      .wait(
-        '"amplify publish" will build all your local backend and frontend resources (if you have hosting category added) and provision it in the cloud',
-      )
+      .wait('"amplify publish" will build all your local backend and frontend resources')
       .run((err: Error) => {
         if (!err) {
           resolve();
@@ -79,10 +77,8 @@ export function addApiWithBlankSchema(cwd: string, opts: Partial<AddApiOptions &
       .sendKeyDown(2)
       .sendCarriageReturn()
       .wait('Do you want to edit the schema now?')
-      .sendLine('n')
-      .wait(
-        '"amplify publish" will build all your local backend and frontend resources (if you have hosting category added) and provision it in the cloud',
-      )
+      .sendConfirmNo()
+      .wait('"amplify publish" will build all your local backend and frontend resources')
       .sendEof()
       .run((err: Error) => {
         if (!err) {
@@ -112,10 +108,8 @@ export function addApiWithBlankSchemaAndConflictDetection(cwd: string) {
       .sendKeyDown(2)
       .sendCarriageReturn()
       .wait('Do you want to edit the schema now?')
-      .sendLine('n')
-      .wait(
-        '"amplify publish" will build all your local backend and frontend resources (if you have hosting category added) and provision it in the cloud',
-      )
+      .sendConfirmNo()
+      .wait('"amplify publish" will build all your local backend and frontend resources')
       .run((err: Error) => {
         if (!err) {
           resolve();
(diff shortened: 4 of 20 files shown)


.codebuild/scripts/e2e_test_windows.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# set exit on error to true
+set -e
+
+source ./shared-scripts.sh && _setupE2ETestsWindows
+codebuild-breakpoint
```
dpilch, line 7: This is keeping the same behavior as E2E tests on linux. I'm not sure `codebuild-breakpoint` is strictly needed on either.

packages/amplify-codegen-e2e-core/src/categories/api.ts
```diff
@@ -49,9 +49,7 @@ export function addApiWithoutSchema(cwd: string, opts: Partial<AddApiOptions & {
       .sendCarriageReturn()
       .wait('Do you want to edit the schema now?')
       .sendConfirmNo()
-      .wait(
-        '"amplify publish" will build all your local backend and frontend resources (if you have hosting category added) and provision it in the cloud',
-      )
+      .wait('"amplify publish" will build all your local backend and frontend resources')
```
dpilch, line 52: Strange error.

```
● codegen remove tests - iOS › Does not delete files during codegen remove
--
351 |  
352 | Expected value   "\"amplify publish\" will build all your local backend and frontend resources (if you have hosting category added) and provision it in the cloud"
353 | Received:
354 | "ision it in the cloud"
355 |  
356 | Message:
357 | expected "ision it in the cloud" to contain "\"amplify publish\" will build all your local backend and frontend resources (if you have hosting category added) and provision it in the cloud"
358 |  
359 | Difference:
360 |  
361 | - Expected
362 | + Received
363 |  
364 | - "amplify publish" will build all your local backend and frontend resources (if you have hosting category added) and provision it in the cloud
365 | + ision it in the cloud
```

The message is correct when running manually. Maybe there is a maximum length? The "v" in "provision" is the 120th character.

This change was also made on the CLI E2E when enabling windows.
https://github.com/aws-amplify/amplify-cli/commit/97c6de42f906d56a5fd0d78c4827110508955b93

shared-scripts.sh
```diff
@@ -235,18 +244,32 @@ function _setupE2ETestsLinux {
...
     retry runE2eTest
 }
 
+function _runE2ETestsWindows {
+    echo "RUN E2E Tests Windows"
+    retry runE2eTest
+}
+
 function _scanArtifacts {
-    if ! yarn ts-node .codebuild/scripts/scan_artifacts.ts; then
```
dpilch, line 249: Get an error when using yarn over npx/npm. It looks like there are some workarounds, but it is easier to just use npx/npm here because there will be no functional difference to the scripts.

```
error Couldn't find the binary C:\codebuild\tmp\output\src2004\src\github.com\aws-amplify\amplify-codegen\node_modules\.bin\ts-node
```

.codebuild/scripts/local_publish_helpers.sh
```diff
@@ -7,7 +7,7 @@ function startLocalRegistry {
     # Start local registry
     tmp_registry_log="$(mktemp)"
     echo "Registry output file: $tmp_registry_log"
-    (cd && nohup npx ${VERDACCIO_PACKAGE:-$default_verdaccio_package} -c $1 &>$tmp_registry_log &)
+    nohup npx ${VERDACCIO_PACKAGE:-$default_verdaccio_package} -c $1 &>$tmp_registry_log &
```
dpilch, line 10: The `cd` is unnecessary here and caused some issues on Windows.

packages/amplify-codegen-e2e-core/package.json
```diff
@@ -32,7 +32,7 @@
     "jest-circus": "^27.5.1",
     "jest-environment-node": "^26.6.2",
     "lodash": "^4.17.19",
-    "node-pty": "beta",
+    "node-pty": "1.0.0",
```
dpilch, line 35: This might not be necessary for this change. I tried this when troubleshooting various issues. However, this is probably something we should do regardless so that our version is stable. I can try removing this and adding it to a separate change if it is desired.

scripts/split-e2e-tests.ts
```diff
@@ -71,7 +71,11 @@ type CandidateJob = {
   runSolo: boolean;
 };
 const createJob = (os: OS_TYPE, jobIdx: number, runSolo: boolean = false): CandidateJob => {
-  const region = AWS_REGIONS_TO_RUN_TESTS[jobIdx % AWS_REGIONS_TO_RUN_TESTS.length];
+  // The bash terminal for Windows on CodeBuild is non-interactive.
+  // amplify-configure will always choose us-east-1 for the region.
+  // This will set all Windows jobs to use us-east-1 to avoid region mismatch.
+  // We will come back to this later to properly fix the testing issue.
+  const region = os === 'w' ? 'us-east-1' : AWS_REGIONS_TO_RUN_TESTS[jobIdx % AWS_REGIONS_TO_RUN_TESTS.length];
```
dpilch, line 78: Another weird thing, this only seems to affect the `amplify configure` step. On command like `codegen add` the selection is set correctly.

alharris-at (approved):

.codebuild/scripts/e2e_test_windows.sh
phani-srikar: This is used when you want to run the tests in debug mode. I think the script is called `yarn-e2e-debug`.

packages/amplify-codegen-e2e-core/src/categories/api.ts
phani-srikar: Alternatively, you can wrap the message with `*/` regex to ignore subsequent parts. Just FYI.

packages/amplify-codegen-e2e-tests/src/__tests__/graphql-documents-generator.test.ts
```diff
@@ -31,9 +32,10 @@ describe('GraphQL documents generator e2e tests', () => {
   });
 
   const schemaFileName = 'schema.graphql';
-  it('generates valid GraphQL documents for given schema', async () => {
+  // skip cypress test on windows
+  (isWindows() ? it.skip : it)('generates valid GraphQL documents for given schema', async () => {
```
phani-srikar, line 36: Do we want to run these on windows as well eventually? If yes, we should have a task for it.

scripts/split-e2e-tests.ts
phani-srikar: We might need to look into how we're sending the region selection command interactively and make it for both platforms. Running in `us-east-1` is fine for now.

phani-srikar (approved):

scripts/split-e2e-tests.ts
dpilch: It is using j/k to navigate the list. I tried with up/down arrow as well and ran into the same issue. I've created a task to investigate further.

packages/amplify-codegen-e2e-tests/src/__tests__/graphql-documents-generator.test.ts
dpilch: I've created tasks for the 2 disabled tests and the region issue.

PR #793, opened 2024-03-06
feat: set correct association with references

dpilch:
#### Description of changes

Update associations to use custom field names when using `references`. Previously associations to used the implicit names when `references` were set.

See https://github.com/aws-amplify/amplify-codegen/pull/806 for spec.

N/A

#### Issue #, if available

N/A

* Unit tests
* E2E tests
* Add schema with references to test build apps: https://github.com/aws-amplify/amplify-codegen/pull/809
  * Further validation for codegen will be done before merging to main
  * The test schemas have been added to the library on the gen2 branch: https://github.com/aws-amplify/amplify-category-api/pull/2442

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Changes are tested on windows. Some Node functions (such as `path`) behave differently on windows.


packages/appsync-modelgen-plugin/src/scalars/supported-directives.ts
```diff
@@ -139,10 +139,12 @@ export const directives = /* GraphQL */ `
   directive @primaryKey(sortKeyFields: [String!], queryField: String) on FIELD_DEFINITION
   directive @index(name: String!, sortKeyFields: [String], queryField: String) on FIELD_DEFINITION
 
-  directive @hasOne(fields: [String!]) on FIELD_DEFINITION
-  directive @hasMany(indexName: String, fields: [String], limit: Int = 100) on FIELD_DEFINITION
-  directive @belongsTo(fields: [String]) on FIELD_DEFINITION
+  directive @hasOne(fields: [String!], references: [String!]) on FIELD_DEFINITION
+  directive @hasMany(indexName: String, fields: [String], limit: Int = 100, references: [String!]) on FIELD_DEFINITION
```
atierian, line 143: Does ordering matter here? The `HasManyTransformer` uses a directive definition with different ordering. [[source](https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-graphql-relational-transformer/src/graphql-has-many-transformer.ts#L50)]
`directive @hasMany(indexName: String, fields: [String!], references: [String!], limit: Int = ${defaultLimit}) on FIELD_DEFINITION`

dpilch: I don't believe so. I can't find any documentation on it though. I'll update to be safe. 

It won't matter after https://github.com/aws-amplify/amplify-category-api/pull/2342.

atierian (approved):

packages/appsync-modelgen-plugin/src/scalars/supported-directives.ts
palpatim: Directive ordering is significant only when they're actually being applied. Order of definition is not significant.

packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-model-introspection-visitor.test.ts
```diff
@@ -503,7 +503,262 @@ describe('custom fields', () => {
...
+      }
+    `;
+
+    const visitor: AppSyncModelIntrospectionVisitor = getVisitor(schema);
+    expect(visitor.generate()).toMatchSnapshot();
+  });
+
+  test('sets the association to the references field for hasOne and hasMany', () => {
+    const schema = /* GraphQL */ `
+      type Primary @model @auth(rules: [{ allow: public, operations: [read] }, { allow: owner }]) {
```
palpatim, line 552: remove `@auth` rules?

packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-visitor.test.ts
```diff
@@ -1410,4 +1410,37 @@ describe('AppSyncModelVisitor', () => {
       expect(interfaces).toMatchSnapshot();
     });
   })
+
+  describe('references field on hasOne, hasMany, and belongsTo directive', () => {
+    test('converts string argument to list of string', () => {
+      const schema = /* GraphQL */ `
+        type Primary @model @auth(rules: [{ allow: public, operations: [read] }, { allow: owner }]) {
```
palpatim, line 1417: remove `@auth`

packages/appsync-modelgen-plugin/src/utils/process-belongs-to.ts
```diff
@@ -55,9 +60,10 @@ export function processBelongsToConnection(
   return {
     kind: CodeGenConnectionType.BELONGS_TO,
     connectedModel: otherSide,
-    isConnectingFieldAutoCreated,
+    isConnectingFieldAutoCreated: false,
     targetName: targetNames[0],
     targetNames,
+    reference,
```
palpatim, line 66: naming nit for the variable and the `CodeGenFieldConnection` type: `useReferences` or `isUsingReferences`?

packages/appsync-modelgen-plugin/src/utils/process-connections-v2.ts
```diff
@@ -20,19 +20,51 @@ export function getConnectedFieldV2(
     throw new Error(`The ${field.name} on model ${model.name} is not connected`);
   }
 
+  const references = connectionInfo.arguments.references;
   if (connectionInfo.name === 'belongsTo') {
     let connectedFieldsBelongsTo = getBelongsToConnectedFields(model, connectedModel);
+    if (references) {
+      if (connectionInfo) {
```
palpatim, line 27: `connectionInfo` can never be falsy since you're checking for `connectionInfo.name` on L24 above

packages/appsync-modelgen-plugin/src/utils/process-connections.ts
```diff
@@ -17,14 +17,14 @@ export type CodeGenFieldConnectionBelongsTo = CodeGenConnectionTypeBase & {
   kind: CodeGenConnectionType.BELONGS_TO;
   targetName: string; // Legacy field remained for backward compatability
   targetNames: string[]; // New attribute for v2 custom pk support
-
+  reference?: boolean;
```
palpatim, line 20: see naming nit above.

Do we need to modify the publicly exported type? If so, we need to run this change by studio, console, runtime folks to make sure the new field isn't going to break anybody. It seems like the only reason you need it is for internal processing, so using a dedicated internal type would prevent this implementation detail from leaking into your public abstraction.

If we for some reason do need the consumers to know about references, then that's a whole other conversation we need to have. :)

dpilch: These types are for the intermediate representation and are not exported outside of the package. i.e. No codegen outputs will contain this field.

This boolean is only necessary for this case: https://github.com/aws-amplify/amplify-codegen/blob/b2780a920ca7295a3098bda4c7ed50be7686d9ea/packages/appsync-modelgen-plugin/src/visitors/appsync-visitor.ts#L1193-L1212

It would be possible to re-process the relationship from here, but I think it would become much more complicated.

palpatim (approved):

atierian (approved):

PR #807, opened 2024-04-09
fix: use standalone ajv validation for model introspection schema

AaronZyLee:
#### Description of changes

The direct `Ajv` dependency uses a  a browser-unsafe eval which will cause a CSP(Content Security Policy) error during runtime. To resolve the issue, the following changes are made:

- `Ajv` is removed from dependency but added in dev dependency
- Use the standalone validation code generated by the `Ajv` library as suggested by the documentation to avoid the unsafe-val method to be used in browser.
  - https://ajv.js.org/standalone.html

Unit & E2E tests

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/appsync-modelgen-plugin/src/validate-cjs.js
AaronZyLee, line 1: This file is auto-generated by the Ajv library

packages/appsync-modelgen-plugin/scripts/generateStandaloneValidationFunction.ts
AaronZyLee, line 1: See example code https://ajv.js.org/standalone.html#generating-using-the-js-library

packages/appsync-modelgen-plugin/package.json
```diff
@@ -47,7 +47,8 @@
     "@types/pluralize": "0.0.29",
     "graphql": "^15.5.0",
     "java-ast": "^0.3.0",
-    "ts-json-schema-generator": "1.0.0"
+    "ts-json-schema-generator": "1.0.0",
+    "ajv": "^8.12.0"
```
palpatim, line 51: nit: please alphabetize dependencies so we can keep track of them more easily

packages/appsync-modelgen-plugin/src/visitors/appsync-model-introspection-visitor.ts
```diff
@@ -43,8 +38,8 @@ export class AppSyncModelIntrospectionVisitor<
     );
 
     const modelIntrosepctionSchema = this.generateModelIntrospectionSchema();
-    if (!this.schemaValidator(modelIntrosepctionSchema)) {
-      throw new Error(`Data did not validate against the supplied schema. Underlying errors were ${JSON.stringify(this.schemaValidator.errors)}`);
+    if (!validateJsonFile(modelIntrosepctionSchema)) {
```
palpatim, line 41: If we don't already have them, please add both positive & negative tests that exercise this validation

AaronZyLee: Thanks. I've added some basic test cases for the standalone validator. The complex cases are already covered by the MIS visitor tests.

palpatim (approved):

packages/appsync-modelgen-plugin/scripts/generateStandaloneValidationFunction.ts
```diff
@@ -0,0 +1,14 @@
+import Ajv from 'ajv';
+import modelIntrospectionSchemaDefinition from '../schemas/introspection/1/ModelIntrospectionSchema.json'
```
phani-srikar, line 2: nit for future: Since we're versioning the MIS definition and this script generates the standalone for version 1 of MIS, we can maybe make it more generic so that it accepts the version input and outputs one standalone per version.

phani-srikar (approved):

PR #813, opened 2024-04-18
fix!: use transformer version 2 by default

dpilch:
BREAKING CHANGE: use transformer version 2 by default in graphql-generator package

#### Description of changes

`transformerVersion` parameter had the wrong type. It was boolean instead of number. With the current default `true` the transformer version 1 was being used by default.

This means that gen 2 model generation is using has `transformerVersion` set to 1. Gen 1 is still using the cli.json feature flag which has explicitly set the version.

The parameter is used here: 
1. https://github.com/aws-amplify/amplify-codegen/blob/9f559413621dcc4a4a5158a129e575371c6d8c35/packages/appsync-modelgen-plugin/src/visitors/appsync-swift-visitor.ts#L394
2. https://github.com/aws-amplify/amplify-codegen/blob/9f559413621dcc4a4a5158a129e575371c6d8c35/packages/appsync-modelgen-plugin/src/preset.ts#L325

This is a breaking change for the graphql-generator. This package is intended for internal use only.

We will still need to upgrade the package version in gen 1 and gen 2 CLI.

Gen 1 behavior will not change with this upgrade.

Gen 2 behavior will be breaking, but it is still pre-GA.

* `transformerVersion` type changed to number.
* `transformerVersion` default changed to 2.

#### Issue #, if available

N/A

* Unit tests
* E2E tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/graphql-generator/API.md
+++ b/packages/graphql-generator/API.md
@@ -25,7 +25,7 @@ export type GenerateModelsOptions = {
     generateIndexRules?: boolean;
     emitAuthProvider?: boolean;
     useExperimentalPipelinedTransformer?: boolean;
-    transformerVersion?: boolean;
+    transformerVersion?: number;
     respectPrimaryKeyAttributesOnConnectionField?: boolean;
     improvePluralization?: boolean;
     generateModelsForLazyLoadAndCustomSelectionSet?: boolean;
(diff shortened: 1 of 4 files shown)


palpatim:
Please change the title to conventional commit breaking change format to ensure we get a MV bump on the next release: `fix!: use transformer version 2 by default`

palpatim (approved):
Approved contingent on changing PR title to conform to conventional commit breaking change format (e.g., `fix!: use transformer version 2 by default` or by including a `BREAKING CHANGE:` in the footer

phani-srikar (approved):

PR #869, opened 2024-08-28
feat(appsync-modelgen-plugin): add support for generation route definitions

atierian:
#### Description of changes
- Adds `generations` key to model introspection schema.
- Prevents generation routes (Query fields with `@generation` directive) from populating existing `queries` key in MIS; instead assigns them to the new `generations` key.

#### Codegen Paramaters Changed or Added
N/A

#### Issue #, if available
N/A

#### Description of how you validated changes
- Snapshot testing
- Manual E2E testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed or added](https://github.com/aws-amplify/amplify-codegen/blob/main/CONTRIBUTING.md#tests)
- [ ] ~Relevant documentation is changed or added (and PR referenced)~ Docs PRs are WIP
- [x] Breaking changes to existing customers are released behind a feature flag or major version update
- [ ] ~Changes are tested using sample applications for all relevant platforms (iOS/android/flutter/Javascript) that use the feature added/modified~
- [ ] ~Changes are tested on windows. Some Node functions (such as `path`) behave differently on windows.~
- [x] Changes adhere to the [GraphQL Spec](https://spec.graphql.org/June2018/) and supports the GraphQL types `type`, `input`, `enum`, `interface`, `union` and scalar types.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #872, opened 2024-09-04
feat(conversation): add conversation route handling in mis

atierian:
# Description of changes
- Adds handling of `conversation` directives for model introspection schema.

## A note on breaking changes
I've confirmed with the Console team and tested Data Schema myself that this change doesn't break consumers. Unknown keys, such as `conversations` and `generations` (https://github.com/aws-amplify/amplify-codegen/pull/869) are ignored by both consumers. **However**, a consumer of codegen that:
- uses a version of codegen without knowledge of the `@conversation` or `@generation` directive

and

- generates the model introspection schema based on an Ampilfy SDL GraphQL schema containing a `@conversation` or `@generation` directive

will break because codegen isn't aware of these directives.

For Data Schema, this in unproblematic. Customers won't be able to define conversation or generation routes in their schema without using (transitively) a version of `amplify-graphql-directives` that defines these directives.

For Console, this could be problematic for customers that define conversation or generation routes **and** use data manager in the console. The console team is aware of the impending releases and will work on updating their packages.

`conversations` key within MIS.

#### Issue #, if available
N/A

Automated snapshot and manual E2E tests.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] ~Relevant documentation is changed or added (and PR referenced)~
- [ ] ~Breaking changes to existing customers are released behind a feature flag or major version update~
- [ ] ~Changes are tested using sample applications for all relevant platforms (iOS/android/flutter/Javascript) that use the feature added/modified~
- [ ] ~Changes are tested on windows. Some Node functions (such as `path`) behave differently on windows.~
- [x] Changes adhere to the [GraphQL Spec](https://spec.graphql.org/June2018/) and supports the GraphQL types `type`, `input`, `enum`, `interface`, `union` and scalar types.


packages/appsync-modelgen-plugin/src/utils/process-conversation.ts
```diff
@@ -0,0 +1,198 @@
...
+        ...generateGraphQLOperationMetadata<CodeGenMutation, SchemaMutation>(mutationObj),
+        type: { model: conversationMessageModelName }
+      },
+      subscribe: generateSubscriptionMetadata(routeName, conversationMessageModelName),
+    }
+  };
+}
+
+/**
+ * Generates a conversation model schema.
```
palpatim, line 48: As discussed, let's add a note here that this duplicates the conversation model creation logic in the transformer, and the two should be evolved in parallel. (Future PR is fine)

palpatim (approved):

palpatim (approved):

PR #879, opened 2024-09-17
fix(amplify-codegen): support multiple indexes on the same field in introspection schema

sundersc:
#### Description of changes
Fixes a bug where multiple indexes on a field is not correctly represented on the generated model introspection schema.

No.

#### Issue #, if available

Fixes https://github.com/aws-amplify/amplify-category-api/issues/2815

- Unit tests
- Manual test

> [!NOTE]  
> **Manual testing:** I've tested the fix on a sample Gen2 app using NPM overrides pointing to the local packaged module.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Changes adhere to the [GraphQL Spec](https://spec.graphql.org/June2018/) and supports the GraphQL types `type`, `input`, `enum`, `interface`, `union` and scalar types.

### DIFF
--- a/packages/appsync-modelgen-plugin/src/utils/process-index.ts
+++ b/packages/appsync-modelgen-plugin/src/utils/process-index.ts
@@ -1,5 +1,5 @@
 import { CodeGenDirective, CodeGenModel } from '../visitors/appsync-visitor';
-import { getDirective } from './fieldUtils';
+import { getDirectives } from './fieldUtils';
 import pluralize from 'pluralize';
 import { toLower, toUpper } from './stringUtils';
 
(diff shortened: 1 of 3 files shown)


packages/appsync-modelgen-plugin/src/utils/process-index.ts
```diff
@@ -9,21 +9,27 @@ import { toLower, toUpper } from './stringUtils';
...
+    if (!indexDirectives) {
       return acc;
     }
-    return { ...acc, [field.name]: indexDirective };
-  }, {} as Record<string, CodeGenDirective>);
+    return { ...acc, [field.name]: indexDirectives };
+  }, {} as Record<string, CodeGenDirective[]>);
+
+  const keyList: CodeGenDirective[] = [];
+  Object.entries(indexMap).forEach(([fieldName, directives]) => {
```
dpilch, line 20: nit: Why change this from a map to a forEach?

dpilch (approved):

packages/appsync-modelgen-plugin/src/utils/process-index.ts
sundersc: `.map` returns a new array, whereas `.forEach` doesn't return anything. Since the refactored code doesn't rely on the returned array, `.forEach` is a better fit here.

palpatim (approved):

PR #904, opened 2024-10-31
feat: added canary yml file

tejas2008:
#### Description of changes
Added yml file for canary workflow in .codebuild

N/A

#### Issue #, if available
N/A

#### Checklist

- [x] PR description included

### DIFF
--- a/.codebuild/canary_workflow.yml
+++ b/.codebuild/canary_workflow.yml
@@ -1,4 +1,4 @@
- version: 0.2
+version: 0.2
 env:
   shell: bash
   compute-type: BUILD_GENERAL1_LARGE
@@ -28,7 +28,6 @@ batch:
         variables:
           TEST_SUITE: src/__tests__/build-app-swift.test.ts
           CLI_REGION: us-east-1
-      debug-session: true
       depend-on:
         - publish_to_local_registry
     - identifier: build_app_android_us_east_1
@@ -38,7 +37,6 @@ batch:
         variables:
           TEST_SUITE: src/__tests__/build-app-android.test.ts
           CLI_REGION: us-east-1
-      debug-session: true
       depend-on:
         - publish_to_local_registry
     - identifier: build_app_ts_us_east_1
@@ -48,7 +46,6 @@ batch:
         variables:
           TEST_SUITE: src/__tests__/build-app-ts.test.ts
           CLI_REGION: us-east-1
-      debug-session: true
       depend-on:
         - publish_to_local_registry
     - identifier: cleanup_e2e_resources
--- a/.codebuild/canary_workflow.yml
+++ b/.codebuild/canary_workflow.yml
@@ -31,23 +31,23 @@ batch:
       debug-session: true
       depend-on:
         - publish_to_local_registry
-    - identifier: build_app_android
+    - identifier: build_app_android_us_east_1
       buildspec: .codebuild/run_android_modelgen_e2e_test.yml
       env:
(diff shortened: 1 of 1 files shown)


.codebuild/canary_workflow.yml
```diff
@@ -0,0 +1,59 @@
...
+      depend-on:
+        - build_linux
+    - identifier: build_app_swift_us_east_1
+      buildspec: .codebuild/run_ios_modelgen_e2e_test.yml
+      env:
+        compute-type: BUILD_GENERAL1_LARGE
+        variables:
+          TEST_SUITE: src/__tests__/build-app-swift.test.ts
+          CLI_REGION: us-east-1
+      debug-session: true
```
atierian, line 31: here and below, why `debug-session: true`?

.codebuild/canary_workflow.yml
```diff
@@ -0,0 +1,59 @@
...
+          CLI_REGION: ap-southeast-1
+      debug-session: true
+      depend-on:
+        - publish_to_local_registry
+    - identifier: cleanup_e2e_resources
+      buildspec: .codebuild/cleanup_e2e_resources.yml
+      env:
+        compute-type: BUILD_GENERAL1_MEDIUM
+      depend-on:
+        - publish_to_local_registry
```
atierian, line 59: This will start running `cleanup_e2e_resources` while the other jobs are running (once `publish_to_local_registry` completes), right? Is that intentional?

.codebuild/canary_workflow.yml
```diff
@@ -0,0 +1,59 @@
...
+      buildspec: .codebuild/run_ios_modelgen_e2e_test.yml
+      env:
+        compute-type: BUILD_GENERAL1_LARGE
+        variables:
+          TEST_SUITE: src/__tests__/build-app-swift.test.ts
+          CLI_REGION: us-east-1
+      debug-session: true
+      depend-on:
+        - publish_to_local_registry
+    - identifier: build_app_android
```
phani-srikar, line 34: nit: keep the naming consistent b/w `build_app_swift_us_east_1` and `build_app_android`

phani-srikar (commented):
Any reason for choosing particular region for each canary? We could either keep the region consistent or have the canary run in all supported regions (like in API repo canaries).

.codebuild/canary_workflow.yml
tejas2008: My thought was if any test fails atleast the `cleanup_e2e_resources` run but I guess I ran the build again and it executes it before which is expected and it will leave the tests after it so I will update its dependency to the android test.

.codebuild/canary_workflow.yml
tejas2008: In the `codegen e2e workflow` the `debug -session:  true` was present so I replicated the same Do you recommend removing it?

.codebuild/canary_workflow.yml
tejas2008: Yes will update the naming nit and regions accordingly

phani-srikar (dismissed):

.codebuild/canary_workflow.yml
atierian: Yes because canaries are run out of band from typical CI workflows triggered on PR / commit.

Including the flag here doesn't provide value and is surprising / confusing to see in a canary workflow.

atierian (approved):

phani-srikar (approved):

PR #931, opened 2025-02-10
fix: stop pascal casing enums for MIS

atierian:
#### Description of changes

Codegen currently pascal cases enums defined in the GraphQL schema.

Gen 2 doesn't pascal case top-level enums:
```ts
const schema = a.schema({
  foo: a.enum(["a1", "b1"]),
});
```
generates:
```gql
enum foo {
  a1
  b1
}
```

This naming discrepancy hasn't caused problems yet because:
1. Type names for return types aren’t included in GraphQL operations.
2. The only way an enum can be used as an input today is through a generated model query. For example:
```ts
const schema = a.schema({
  foo: a.enum(["a1", "b1"]),
  Todo: a.model({
    status: a.boolean(),
    test: a.ref('foo')
  }),
});
```
But we create a wrapper input type for all model generated queries. So the actual GraphQL mutation here uses CreateTodoInput + Variables and never mentions the `foo` type.
```js
{
  "query": "mutation ($input: CreateTodoInput!) {
    createTodo(input: $input) {
      id
      status
      test
      createdAt
      updatedAt
    }
  }",
  "variables": {
    "input": {
        "status": true,
        "test": "b1"
    }
  }
}
```

This changes with the custom type / enum input for custom operation work in progress.
```ts
const schema = a.schema({
  foo: a.enum(["a1", "b1"]),
  fcnCall: a
    .query()
    .arguments({
      e3: a.ref("foo"),
    }),
});
```
will generate a GraphQL query like:
```js
{
  "query": "query ($e3: Foo) {
    fcnCall(e3: $e3)
  }",
  "variables": {
    "e3": "a1"
  }
}
```
because there isn’t a wrapper input type.

> [!NOTE]
> This change only applies to the model introspection schema.
> Codegenerated enums are still pascal cased because changing that will cause a breaking change if the schema defined enum casing doesn't match. Non-MIS consuming clients (e.g. native codegen) doesn't support custom operations, so the feature motivating this change doesn't apply.

#### Issue #, if available
N/A

- [E2E test run](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/594813022831/projects/amplify-codegen-e2e-workflow/batch/amplify-codegen-e2e-workflow:1b6f242b-973e-4946-9a1f-f65b607965f6?region=us-east-1)

> [!NOTE]
> Two Gen 2 related E2Es are failing in this run, but they have been failing for a long time and the failures are unrelated. I'm unsure of when the change causing these failures was introduced. The only way I've been able to repro the failure locally is by running `yarn e2e ...` instead of `npm e2e ...`, which results in both yarn.lock and package-lock.json files being generated in the project root. This in turn causes an error when deploying and mimics the observed failures in CloudBuild runs.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] E2E test run linked
- [ ] ~Relevant documentation is changed or added (and PR referenced)~
- [ ] ~Breaking changes to existing customers are released behind a feature flag or major version update~
- [ ] ~Changes are tested using sample applications for all relevant platforms (iOS/android/flutter/Javascript) that use the feature added/modified~
- [ ] ~Changes are tested on windows. Some Node functions (such as `path`) behave differently on windows.~
- [x] Changes adhere to the [GraphQL Spec](https://spec.graphql.org/June2018/) and supports the GraphQL types `type`, `input`, `enum`, `interface`, `union` and scalar types.

### DIFF
--- a/packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-model-introspection-visitor.test.ts
+++ b/packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-model-introspection-visitor.test.ts
@@ -115,6 +115,10 @@ describe('Model Introspection Visitor', () => {
     interface SimpleInterface {
       firstName: String!
     }
+    enum camel {
+      bactrian
+      dromedary
+    }
     union SimpleUnion = SimpleModel | SimpleEnum | SimpleNonModelType | SimpleInput | SimpleInterface
   `;
   const visitor: AppSyncModelIntrospectionVisitor = getVisitor(schema);
@@ -142,6 +146,10 @@ describe('Model Introspection Visitor', () => {
     it('should return interface type for Interface', () => {
       expect((visitor as any).getType('SimpleInterface')).toEqual({ interface: 'SimpleInterface' });
     });
+
+    it('should not pascal case enum', () => {
+      expect((visitor as any).getType('camel')).toEqual({ enum: 'camel' })
+    })
 
     it('should throw error for unknown type', () => {
       expect(() => (visitor as any).getType('unknown')).toThrowError('Unknown type');
--- a/packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-model-introspection-visitor.test.ts
+++ b/packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-model-introspection-visitor.test.ts
@@ -115,7 +115,7 @@ describe('Model Introspection Visitor', () => {
     interface SimpleInterface {
       firstName: String!
     }
-    enum camel {
+    enum camelCase {
       bactrian
       dromedary
     }
@@ -148,7 +148,7 @@ describe('Model Introspection Visitor', () => {
     });
 
     it('should not pascal case enum', () => {
-      expect((visitor as any).getType('camel')).toEqual({ enum: 'camel' })
+      expect((visitor as any).getType('camelCase')).toEqual({ enum: 'camelCase' })
     })
 
     it('should throw error for unknown type', () => {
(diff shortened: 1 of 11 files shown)


packages/appsync-modelgen-plugin/src/visitors/appsync-swift-visitor.ts
```diff
@@ -305,7 +305,7 @@ export class AppSyncSwiftVisitor<
         .asKind('enum')
         .access('public')
         .withProtocols(['String', 'EnumPersistable'])
-        .withName(this.getEnumName(enumValue));
+        .withName(pascalCase(this.getEnumName(enumValue)));
```
dpilch, line 308: Is there a swift test case?

packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-model-introspection-visitor.test.ts
```diff
@@ -143,6 +147,10 @@ describe('Model Introspection Visitor', () => {
       expect((visitor as any).getType('SimpleInterface')).toEqual({ interface: 'SimpleInterface' });
     });
 
+    it('should not pascal case enum', () => {
+      expect((visitor as any).getType('camel')).toEqual({ enum: 'camel' })
```
dpilch, line 151: Can we test with a two word name like `camelCase` to ensure that `Case` is still capitalized?

packages/appsync-modelgen-plugin/src/visitors/appsync-swift-visitor.ts
atierian: Yes - [should generate an enum for enum type](https://github.com/aws-amplify/amplify-codegen/blob/b8df7e6405b7f9888bcbb632702fa9fe43fff980/packages/appsync-modelgen-plugin/src/__tests__/visitors/appsync-swift-visitor.test.ts#L161-L189).

dpilch (approved):

zxl629 (approved):

PR #934, opened 2025-02-12
feat: add regionalized build spec for three canaries

Siqi-Shan:
#### Description of changes

Added regionalized `.yml` CodeBuild spec files for three app builds (`ts`, `swift`, `android`) canaries.

These build spec files are auto generated by running `yarn split-canary-tests`, with regions sourced from [Amplify supported regions](https://docs.aws.amazon.com/general/latest/gr/amplify.html).

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.codebuild/run_regionalized_android_modelgen_e2e_test.yml
+++ b/.codebuild/run_regionalized_android_modelgen_e2e_test.yml
@@ -1,0 +1,35 @@
+version: 0.2
+env:
+  shell: bash
+  variables:
+    AMPLIFY_DIR: /root/.npm-global/lib/node_modules/@aws-amplify/cli-internal/bin
+    AMPLIFY_PATH: /root/.npm-global/lib/node_modules/@aws-amplify/cli-internal/bin/amplify
+    CI: true
+    CODEBUILD: true
+    NODE_OPTIONS: --max-old-space-size=8096
+
+phases:
+  install:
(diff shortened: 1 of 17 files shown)


palpatim (changes requested):
Can we add a script to generate these so it's easier to manage if we decide to make a change to the canary invocation, or add a new region? Right now a change would require a bunch of error-prone manual updates.

.codebuild/run_regionalized_android_modelgen_e2e_test.yml
```diff
@@ -0,0 +1,35 @@
...
+
+phases:
+  install:
+    commands:
+      - sudo apt update
+      - yes | sudo apt install android-sdk
+      - export ANDROID_HOME=/usr/lib/android-sdk
+      - yes | sudo apt install sdkmanager
+      # Review SDK licenses
+      - yes | sudo sdkmanager --licenses
```
palpatim, line 19: I thought we were removing these to the test code? Or is that coming in another PR?

Siqi-Shan: My bad I forgot to pull changes from `main`. This branch was created before last PR

palpatim (approved):

bobbyu99 (approved):

PR #936, opened 2025-02-15
fix: opt in region test to correct account

Siqi-Shan:
#### Description of changes

Revised `useChildAccountCredentials` in `shared-scripts.sh` to let correct child account picked up test with opt-in region.

Following changes are also included:
- Fix current script to generate CodeBuild canary build specs, so correct iOS build spec will be generated.
- Added "unassume credential" step after every Android canary test.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] E2E test run linked


scripts/e2e-test-regions.json
palpatim, line 1: Do we have a comment in the relevant other locations that this is another place we need to keep the region list in sync?

shared-scripts.sh
```diff
@@ -339,10 +339,26 @@ function _unassumeTestAccountCredentials {
...
+                    pick_acct=$child_acct
+                    break
+                fi
+            done
+        else
+            pick_acct=$(echo $child_accts | jq -cr ".[$RANDOM % $org_size]")
+        fi
         session_id=$((1 + $RANDOM % 10000))
         if [[ -z "$pick_acct" || -z "$session_id" ]]; then
           echo "Unable to find a child account. Falling back to parent AWS account"
```
palpatim, line 364: Is this the right way to respond to this failure? If our child accounts are out of sync shouldn't we know about it rather than swallowing it in an echo message that we're unlikely to ever see?

Siqi-Shan: Used same logic of category api's script here when unable to find a child account: https://github.com/aws-amplify/amplify-category-api/blob/main/shared-scripts.sh#L421

We could do `exit 1` here, but AFAIK this branch is very rarely (or never) triggered. If we have concerns I propose we defer that change to a separate PR for better clarity and discussion

scripts/e2e-test-regions.json
Siqi-Shan: Yes, just remembered this file is duplicated with a prev added `support-test-regions.json` file. Removed the old one and updated related comments

.codebuild/ios_canary_workflow.yml
```diff
@@ -27,7 +27,7 @@ batch:
       env:
         compute-type: BUILD_GENERAL1_LARGE
         variables:
-          TEST_SUITE: src/__tests__/build-app-ios.test.ts
+          TEST_SUITE: src/__tests__/build-app-swift.test.ts
```
dpilch, line 30: This was just a typo before?

shared-scripts.sh
palpatim: My concern is that we're now doing actual manipulation of the account, as opposed to a passthrough. We've increased the surface area for errors, but are maintaining the same error strategy that assumes that we won't have an error.

.codebuild/ios_canary_workflow.yml
Siqi-Shan: Yes, I wrote a script in prev PR to auto generate these canary build specs, and made a type for the test source name

shared-scripts.sh
Siqi-Shan: Oh I got your points here. Will update error messages here (and in api repo) so an actual error will be thrown rather than just `echo`. We don't have particular error strategy for this in api repo as well. I'll shortly cut another PR to sync error handling logic

palpatim (approved):

dpilch (approved):

PR #937, opened 2025-02-18
fix: add script to use node 18.20

bobbyu99:
#### Description of changes

`l_graphql_generator_gen2` and `w_graphql_generator_gen2` were failing because CodeBuild was using an older Node version. This PR adds logic to install Node 18.20 for both Windows and Linux environments.

After upgrading to Node 18.20, on Windows, the execution of `npx.cmd` was failing with an `EINVAL` error when calling `spawnSync` in the test. This error is due to a recent [Node.js security update (April 10, 2024)](https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2#command-injection-via-args-parameter-of-child_processspawn-without-shell-option-enabled-on-windows-cve-2024-27980---high). The update enforces that batch files (like `.cmd`) must be run with `shell: true` to prevent potential command injection vulnerabilities. Without this option, Node now errors with `EINVAL` when trying to execute these files directly. By adding `shell: true` to the spawn options, we ensure that the batch file is executed through the shell, which resolves the error and allows the tests to locate and run `ampx` as expected.

[e2e](https://tiny.amazon.com/g6reutmn/IsenLink)
- Note that 2 other e2e are failing as @Siqi-Shan's fix is not merged yet. 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] E2E test run linked
- [ ] Changes adhere to the [GraphQL Spec](https://spec.graphql.org/June2018/) and supports the GraphQL types `type`, `input`, `enum`, `interface`, `union` and scalar types.

### DIFF
--- a/packages/amplify-codegen-e2e-tests/src/gen2-codegen-tests-base/commands.ts
+++ b/packages/amplify-codegen-e2e-tests/src/gen2-codegen-tests-base/commands.ts
@@ -7,57 +7,18 @@ import { spawnSync } from 'child_process';
  * Retrieve the path to the `npx` executable for interacting with the amplify gen2 cli.
  * @returns the local `npx` executable path.
  */
-const getNpxPath = (): string => {
-  console.log(`DEBUG: process.platform is: ${process.platform}`);
-  if (process.platform === 'win32') {
-    const scriptRunnerPath = getScriptRunnerPath();
-    console.log(`DEBUG: getScriptRunnerPath() returned: ${scriptRunnerPath}`);
-    const npxPath = scriptRunnerPath.replace('node.exe', 'npx.cmd');
-    console.log(`DEBUG: Resolved npxPath for Windows: ${npxPath}`);
-    return npxPath;
-  } else {
-    console.log(`DEBUG: Non-Windows platform; using 'npx'`);
-    return 'npx';
-  }
-};
+const getNpxPath = (): string => (process.platform === 'win32' ? getScriptRunnerPath().replace('node.exe', 'npx.cmd') : 'npx');
 
 /**
  * Retrieve the path to the `ampx` executable for interacting with the amplify gen2 cli.
  * @param cwd current working directory
  * @returns the local `ampx` executable path
+ *
+ * Note:
+ * Use shell: true to properly execute the .cmd file on Windows
  */
-const getAmpxPath = (cwd: string): string => {
-  // Retrieve the path to npx
-  const npxPath = getNpxPath();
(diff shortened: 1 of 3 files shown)


shared-scripts.sh
```diff
@@ -301,11 +302,15 @@ function _runE2ETestsWindows {
 
 function _runGen2E2ETestsLinux {
     echo "RUN Gen2 E2E Tests Linux"
+    echo "Setup Node Version"
+    _setupNodeVersionLinux $AMPLIFY_NODE_VERSION
```
dpilch, line 306: With this we will be building and testing with different versions of Node. Would it be possible to install the node version at the build stage so they are consistent?

packages/amplify-codegen-e2e-tests/src/gen2-codegen-tests-base/commands.ts
```diff
@@ -8,13 +8,17 @@ import { spawnSync } from 'child_process';
...
  * Retrieve the path to the `ampx` executable for interacting with the amplify gen2 cli.
  * @param cwd current working directory
  * @returns the local `ampx` executable path
+ *
+ * Note:
+ * Use shell: true to properly execute the .cmd file on Windows
  */
 const getAmpxPath = (cwd: string): string =>
-  spawnSync(getNpxPath(), ['which', 'ampx'], { cwd, env: process.env, stdio: 'pipe' }).stdout.toString().trim();
+  spawnSync(getNpxPath(), ['which', 'ampx'], { cwd, env: process.env, stdio: 'pipe', shell: true }).stdout.toString().trim();
```
palpatim, line 21: Let's do a couple of things here:

1. In your comment on L18, add additional warnings that the invocation uses `shell: true` on Windows, and that command arguments must be sanitized
2. In that comment, add a link to the NodeJS docs about this issue https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2#command-injection-via-args-parameter-of-child_processspawn-without-shell-option-enabled-on-windows-cve-2024-27980---high to give inline explanation why we're using `shell: true`
3. Rather than passing `true` for every platform, scope it to only windows: `shell: process.platform === 'win32'`

palpatim (changes requested):

shared-scripts.sh
bobbyu99: Which build stage are you referring to? `_buildLinux` and `_buildWindows`?

Siqi-Shan (approved):

PR #938, opened 2025-02-18
fix: avoid invalid token error

Siqi-Shan:
#### Description of changes

Added additional handling to invalid token scenarios in cleanup scripts, since some child test accounts do not have opt-in region enabled, and will encounter invalid token error.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] E2E test run linked

### DIFF
--- a/packages/amplify-codegen-e2e-tests/src/cleanup-e2e-resources.ts
+++ b/packages/amplify-codegen-e2e-tests/src/cleanup-e2e-resources.ts
@@ -8,29 +8,24 @@ import fs from 'fs-extra';
 import path from 'path';
 import { deleteS3Bucket, sleep } from '@aws-amplify/amplify-codegen-e2e-core';
 
-// Ensure to update scripts/split-e2e-tests.ts is also updated this gets updated
-const AWS_REGIONS_TO_RUN_TESTS = [
-  'ap-east-1',
-  'ap-northeast-1',
-  'ap-northeast-2',
-  'ap-northeast-3',
-  'ap-south-1',
-  'ap-southeast-1',
-  'ap-southeast-2',
-  'ca-central-1',
-  'eu-central-1',
-  'eu-north-1',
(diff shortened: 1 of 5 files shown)


packages/amplify-codegen-e2e-tests/src/cleanup-e2e-resources.ts
```diff
@@ -97,7 +92,7 @@ type AWSAccountInfo = {
 };
 
 const BUCKET_TEST_REGEX = /test/;
-const IAM_TEST_REGEX = /!RotateE2eAwsToken-e2eTestContextRole|-integtest$|^amplify-|^eu-|^us-|^ap-/;
+const IAM_TEST_REGEX = /!RotateE2eAwsToken-e2eTestContextRole|-integtest|^amplify-|^eu-|^us-|^ap-/;
```
dpilch, line 95: I think this regex is wrong. I'm not aware of any functionality of `!` on its own. I think this would be matching the `!` character literally.

packages/amplify-codegen-e2e-tests/src/cleanup-e2e-resources.ts
```diff
@@ -246,25 +253,36 @@ const getStackDetails = async (stackName: string, account: AWSAccountInfo, regio
...
+          'UPDATE_ROLLBACK_FAILED',
+          'UPDATE_ROLLBACK_COMPLETE',
+          'IMPORT_COMPLETE',
+          'IMPORT_ROLLBACK_FAILED',
+          'IMPORT_ROLLBACK_COMPLETE',
+        ],
+      })
+      .promise();
+  } catch (e) {
+    if (e?.code === 'InvalidClientTokenId') {
```
dpilch, line 275: IIUC this case there shouldn't be anything to cleanup anyway? Can you add to the error message that there isn't any additional cleanup needed?

PR #939, opened 2025-02-19
fix: use correct region in amplify configure

Siqi-Shan:
#### Description of changes

Fixed possible region discrepancy in `amplify configure` due to mismatch between Amplify CLI command dropdown and test script region list.

Plus, removed `ap-east-1` from E2E regions as there is a known difference between Amplify CLI regions and Amplify supported regions. Customer will still be able to deploy app to `ap-east-1`, but this region is not currently supported via `amplify configure`. Having test running in this region will redirect test back to `us-east-1` and failed the Amplify profile assertion in test setup phase.

CI E2Es

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] E2E test run linked

### DIFF
--- a/.codebuild/ts_canary_workflow.yml
+++ b/.codebuild/ts_canary_workflow.yml
@@ -22,17 +22,6 @@ batch:
         compute-type: BUILD_GENERAL1_MEDIUM
       depend-on:
         - build_linux
-    - identifier: build_app_ts_ap_east_1
-      buildspec: .codebuild/run_regionalized_ts_modelgen_e2e_test.yml
-      env:
-        compute-type: BUILD_GENERAL1_LARGE
-        variables:
-          TEST_SUITE: src/__tests__/build-app-ts.test.ts
-          CLI_REGION: ap-east-1
-          CANARY_METRIC_NAME: TsAppBuildCodegenSuccessRate
-          DISABLE_ESLINT_PLUGIN: true
-      depend-on:
-        - publish_to_local_registry
     - identifier: build_app_ts_ap_northeast_1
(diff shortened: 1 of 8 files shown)


bobbyu99 (approved):

packages/amplify-codegen-e2e-core/src/configure/index.ts
```diff
@@ -25,6 +25,21 @@ const defaultProjectSettings = {
   startCmd: '\r',
 };
 
+/**
+ * Supported regions:
```
dpilch, line 29: Can you add a note that the list must be in the same order as what is showing in `amplify configure`.

.codebuild/ts_canary_workflow.yml
```diff
@@ -22,17 +22,6 @@ batch:
         compute-type: BUILD_GENERAL1_MEDIUM
       depend-on:
         - build_linux
-    - identifier: build_app_ts_ap_east_1
```
dpilch, line 25: Will we get canary failure alarms until the CLI adds the missing region to `amplify configure`?

Siqi-Shan: No canary failures, since canary tests will not be blocked by [`testGraphQLConfig`](https://github.com/aws-amplify/amplify-codegen/blob/main/packages/amplify-codegen-e2e-tests/src/codegen-tests-base/test-setup.ts#L40) as regular E2Es (e.g. `add-codegen-*`). But like we discussed `ap-east-1` test will actually run in `us-east-1` and it's not a meaning metric we want to collect

bobbyu99 (approved):

dpilch (approved):

palpatim (approved):

